PR #1, opened 2019-10-10
login flow with airtable

dfangshuo:
Create a login flow using the airtable backend


dfangshuo:
What's in this PR:

1. login flow with airtable, verifying that the username + password (currently plaintext) is correct
2. ensure login state with cookies: if user is logged in, redirect them to home page. If user tries to access home page but is not logged in, bring them to sign in page
3. logout - remove the cookie

PR #2, opened 2019-10-11
login flow with airtable 

dfangshuo:
# Temporary Airlock functionality
What's in this PR:

1. login flow with airtable, verifying that the username + password (currently plaintext) is correct
2. ensure login state with cookies: if user is logged in, redirect them to home page. If user tries to access home page but is not logged in, bring them to sign in page
3. logout - remove the cookie


dfangshuo:
changed PR to remove api key

src/utils/airtableUtils.js
```diff
@@ -0,0 +1,67 @@
...
+                reject(err);
+                return; 
+            }
+            resolve({match: false, found: false});
+        });
+
+    });
+}
+
+var isLoggedIn = email => {
```
aivantg: could you create a helper function to get like....the id of the owner that is logged in? does that exist somewhere?

dfangshuo:
The record id of the person can be obtained from the `getLoggedInUserId` function from `auth.js` @niwong

PR #4, opened 2019-10-15
Asli/admin invite page

asliakalin:
## Admin Invite Page

Added a form as a react component (form.js) where the admins can enter their information and the information of the person they want to invite. The fields are split into 2 categories for the member and the invitee. Member puts their name, surname, username (I don't know if this is something we have I thought it would be unique like ids for indexing referrals in the database later) and their email to maybe get confirmation about their invitation and the good news if their invitee ends up accepting the invitation. Member also puts in the name, lastname and email for the person they are inviting. Also there's a last field about whether the invitation is for a general member or a stakeholder. I assumed the admission process is different for these types.

As of now, the inputs are only collected and outputted as a serialized JSON object. I'll also work on sending the email to both the admin and their invitee and add related data to airtable.

### Screenshots

Screen Shot 2019-10-14 at 5 20 02 PM

CC: @aivantg


src/Invite.js
```diff
@@ -0,0 +1,26 @@
...
+  };
+  
+  onSubmit = fields => { 
+    this.setState({fields});
+  };
+  
+  render() {
+    return (
+      <div className="Invite">
+        <Form onSubmit={fields => this.onSubmit(fields)} />
```
aivantg: I'd go ahead and say that you don't need to encapsulate the form inside a component named `Form`. You can just put all the logic in this component (`Invite.js`)

src/Invite.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import Form from './form';
+
+class Invite extends React.Component {
+  state = {
```
aivantg: I'd prefer if we set our initial state in a constructor. What this looks like is replacing:
```
state = {
  fields: {}
};
```
with 
```
constructor(props) { 
  super(props)
  this.state = {
    fields: {}
  }
}
```

src/form.js
```diff
@@ -0,0 +1,122 @@
+import React from 'react'
+import './form.css';
+
+
+class Form extends React.Component {
```
aivantg: Avoid naming components something vague like `Form`. Instead maybe choose a name like `AdminInviteForm`. Also, this can be combined with `Invite` component as mentioned above. No need for two separate components. 

Let me know if you need help with that!

src/form.js
```diff
@@ -0,0 +1,122 @@
...
+    email: '',
+    username: '',
+    inviteeName: '',
+    inviteeLastName: '',
+    inviteeEmail: '',
+    inviteeType: 'generalMember'
+  }
+}
+  
+change = e => {
```
aivantg: this is a pretty standard function, but I'd still recommend naming it something more verbose like `onChange`

src/Invite.js
asliakalin: fixed for next branch-push

aivantg:
Closing for now.

PR #12, opened 2019-10-25
Nick/general dash bug fix 2

niwong:
## Fix ID-related bug in master.

The ID that was used to get a person record was actually the ID of the User Login table instead of the Person table. 

After pulling master into my user profile page branch, so I could use the login logic, I thought my code was buggy because it would not render the name of the active user. I debugged until I realized that the ID `getLoggedInUserId` returns in auth.js is not the Person ID used to query for a person’s name, but it is instead the ID associated with the User Login table.

This PR has a quick fix for this issue; I am just adding another request in the `componentDidMount` method of the dashboard. I talked it over with Fang, and we plan on working together to probably rewrite the `getLoggedInUserId` function or create a different function that will return the Person ID instead of the User Login ID.

**This GeneralOwnerDashboard.js file should work in an environment that has `Auth.js` located in src/lib/auth.js, but at the git history of this PR, I do not have auth.js.**

### Related PRs

Optional - any related PRs you're waiting on, or PRs that will conflict, etc

### Migrations

Optional - if you added anything to the database through migration(s)

Lol we'll see if we test

### Screenshots

Add screenshots!!! If you'd like

CC: @aivantg


aivantg:
Closing for now, since this PR is out of date

PR #18, opened 2019-11-02
Paypal Payments

dfangshuo:
## Payments w/ Paypal

Making payments using Paypal, currently tested and working using the paypal sandbox.

### !! BREAKING CHANGES 
@niwong 

Moved and modified `secret.js`. Moved `secret.js` from the `src/lib` folder to the root folder. Also updated everywhere that `secret.js` is required, which includes `request.js`. cc-ing @niwong just to keep you in the loop

### Related PRs

- #22 
- #21 

### Migrations

-

The way that you would see this functionality at work is you would go to the subscriber dashboard and click on the yellow Paypal button. This brings you to a sign-in screen. Use the 'Personal' Credentials found on our [Notion page](https://www.notion.so/Payments-23891cdc66964d94b141d8b7604191db). Confirm the payment.

Then, login to https://www.sandbox.paypal.com/. You should still be logged in as the personal account. Logout, and login with the business credentials (found on the same Notion page linked above). The payment should be received and can be seen on the dashboard (last screenshot).

### Screenshots

image

image

image

image

image

CC: @aivantg


dfangshuo:
Also merged in changes from pretty subscriber UI

src/lib/auth.js
```diff
@@ -20,6 +22,7 @@ const table = 'User Login';
 
 const loginUser = (email, passwordHash) => {
   return new Promise((resolve, reject) => {
+    console.log('I WAS CALLED');
```
aivantg: Try and leave these out of your PR's! I know this one will eventually be cleaned up by the airlock PR but just a good thing to keep in mind :)

src/lib/request.js
```diff
@@ -34,27 +36,29 @@ function getRecord(table, id) {
...
-        if (records === null || records.length < 1) {
-          const msg = `No record was retrieved using this ${fieldType}.`;
-          reject(msg);
-        } else {
-          records.forEach(function(record) {
-            resolve(record.fields);
-            return record;
-          });
+        if (records.length < 1) {
+          console.log(`No record was retrieved using this ${fieldType}.`);
```
aivantg: Again, clean up stray log statements if the error case is handled elsewhere

aivantg: but once again, this will almost certainly be resolved when merging with master so no stress

src/screens/general/SubscriberOwnerDashboard.js
```diff
@@ -0,0 +1,213 @@
...
+    const { history } = this.props;
+    const id = getLoggedInUserId();
+    if (!id) {
+      // They shouldn't be able to access this screen
+      history.push('/');
+    }
+  }
+
+  onSuccess(details, data) {
+    /*
```
aivantg: Can these comments be cleaned up? Are they still useful?

src/screens/general/SubscriberOwnerDashboard.js
```diff
@@ -0,0 +1,213 @@
...
+  componentDidMount() {
+    const { history } = this.props;
+    const id = getLoggedInUserId();
+    if (!id) {
+      // They shouldn't be able to access this screen
+      history.push('/');
+    }
+  }
+
+  onSuccess(details, data) {
```
aivantg: I'd like the name of this function to be a little bit more verbose in terms of what it does in relation to the paypal workflow. `onSuccess` is super vague

src/screens/general/SubscriberOwnerDashboard.js
```diff
@@ -0,0 +1,213 @@
...
+        'Payment Create Time': details.create_time,
+        'Payment Update Time': details.update_time
+      }
+    };
+
+    if (id !== orderID) {
+      record.fields.Notes = 'id and order id mismatch';
+    }
+
+    createPayment(record)
```
aivantg: Let's add a TODO to record that this is another important place where retry logic might be important? If paypal goes through but the payment record wasn't created for some reason, it could cause problems

src/screens/general/SubscriberOwnerDashboard.js
aivantg: Maybe link to the documentation instead?

src/screens/general/SubscriberOwnerDashboard.js
aivantg: Also, since this function seems to be a potential general handler for a successful payment, can we maybe move it out to a general payment utility that we can then use for onboarding as well?

src/screens/general/SubscriberOwnerDashboard.js
dfangshuo: There isn't really documentation fo this, which is why I included this in the comments - this information was gotten from looking at the chrome console and manually writing down what the fields are

PR #20, opened 2019-11-08
Create function in `request.js` that creates an Address

dfangshuo:
## FEATURE TITLE

Create function in `request.js` that creates an Address

### Related PRs

-

### Migrations

-

Called the function and verified that it added a record to airtable

### Screenshots

CC: @aivantg


aivantg:
I think this would be a good time to merge Address into Person table. One less object to worry about

PR #26, opened 2019-11-22
WIP: Airlock integration

dfangshuo:
## Integrate Airlock

Integrate Airlock

### Related PRs

N/A

### Migrations

N/A

N/A

### Screenshots

CC: @aivantg

### DIFF
--- a/src/lib/auth.js
+++ b/src/lib/auth.js
@@ -1,5 +1,4 @@
 import Cookies from 'universal-cookie';
-import key from './api_key';
 
 const cookies = new Cookies();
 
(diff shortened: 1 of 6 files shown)


src/lib/auth.js
```diff
@@ -1,64 +1,48 @@
...
   });
 };
 
+const loginUser = async (email, passwordHash) => {
+  try {
+    const { user, token } = await base.login({
+      username: email,
+      password: passwordHash
+    });
+    console.log('SEE ME');
```
aivantg: can we remove these? haha

src/secret.js
```diff
@@ -0,0 +1,5 @@
+const clientId =
+  'AQmImilHhbyAhrxRKFQFLw0NLiyhR10m3yR4snUGC1nQQzLLggGZg6fUfVhiNHi8IgC9stRm4NBVulrd';
```
aivantg: which....key is this? Only airlock key is allowed to be in git I think?

src/lib/auth.js
```diff
@@ -1,64 +1,48 @@
...
+    console.log(user);
+    console.log(token);
```
aivantg: shouldn't this token be saved into the cookie?

arpanpal: The user and token are already stored in localStorage if you need to access in the browser. However, you should always be able to pull from base.getUser() or base.user to get the user, so don't worry about localStorage/cookies. Just know that base.getUser() works!

src/lib/auth.js
```diff
@@ -1,64 +1,48 @@
...
 const cookies = new Cookies();
 
 const BASE_ID = 'appFaOwKhMXrRIQIp';
 
 const Airtable = require('airtable');
 
-const base = new Airtable({ apiKey: key }).base(BASE_ID);
+Airtable.configure({
+  endpointUrl: 'https://airlock-ppsolar.calblueprint.org/',
+  apiKey: 'airlock'
```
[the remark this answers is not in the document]
ethanlee16: Yes, this is safe! We do this to continue causing the Airtable JS client to throw when an API key isn't specified (which is the expected behavior before we forked). More info here: https://www.notion.so/calblueprint/User-Authentication-and-Token-Management-72067d56addf4959a152bfbc53bca089

PR #30, opened 2019-12-04
Address potential inconsistency with Signup

dfangshuo:
## Retries + Rollback

Implemented retries + rollback to rectify potential inconsistencies with creating records on signup

### Related PRs

- https://github.com/calblueprint/peoplepower-web/pull/29

### Screenshots

CC: @aivantg

### DIFF
--- a/src/lib/auth.js
+++ b/src/lib/auth.js
@@ -23,19 +23,20 @@ const USER_LOGIN_TABLE = 'User Login';
 const PERSON_TABLE = 'Person';
 const OWNER_TABLE = 'Owner';
 
-const createPersonOwnerUserLoginRecord = async (
+const DEFAULT_NUM_RETRIES = 3;
+
+const createPersonWithRetries = async (
   email,
-  password,
   phoneNumber,
   fullName,
   street,
-  apt,
   city,
   state,
-  zipcode
+  apt,
+  zipcode,
+  numRetries
 ) => {
   // create a person record without an owner field nor user login field
-  let createdPersonId;
   try {
     const createdPerson = await base(PERSON_TABLE).create([
       {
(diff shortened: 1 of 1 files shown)


src/lib/auth.js
```diff
@@ -51,33 +52,53 @@ const createPersonOwnerUserLoginRecord = async (
...
-    createdPersonId = createdPerson[0].id;
+    return createdPerson[0].id;
   } catch (err) {
-    console.error(err);
-    return false;
+    if (numRetries === 0) {
+      throw err;
+    }
+
+    return createPersonWithRetries(
```
aivantg: recursion in a REAL WORLD PROJECT???? What a concept :P

src/lib/auth.js
```diff
@@ -90,10 +111,112 @@ const createPersonOwnerUserLoginRecord = async (
...
+      createdPersonId,
+      createdOwnerId,
+      email,
+      password,
+      numRetries - 1
+    );
+  }
+};
+
+// TODO: UNTESTED FUNCTION
```
aivantg: Fine with leaving this in here because there are bigger fish to fry, but definitely make sure we come back to this next semester and make sure this logic works

src/lib/auth.js
aivantg: recursion in a REAL WORLD PROJECT, wow never seen that before :P

src/lib/auth.js
```diff
@@ -90,10 +111,112 @@ const createPersonOwnerUserLoginRecord = async (
...
+      createdOwnerId,
+      email,
+      password,
+      numRetries - 1
+    );
+  } catch (err) {
+    console.error(err);
+
+    // if fail to create a User Login record, rollback the created person and owner
+    await rollbackPersonWithRetries(createdPersonId, numRetries);
```
aivantg: I like this! Of course there is the possibility...that that fails...which I guess is the intention of the retry logic...but what if...that fails :'( so many failure scenarios. 

I think this is robust enough for now and not worth chasing any further, but this has me thinking that these rare error cases should maybe _go_ somewhere so that an admin can take care of them in the Airtable. Will note as something to consider in the spring!

PR #42, opened 2019-12-23
[Easy] Clean up constants

dfangshuo:
## Clean up constants

move Airtable table consts + other params into a constants.js file

### Screenshots

CC: @aivantg

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,65 +1,44 @@
-const LOGIN_TOKEN_NAME = 'loginToken';
+const constants = {
+  LOGIN_TOKEN_NAME: 'loginToken',
 
-// AIRTABLE
-const BASE_ID = 'appFaOwKhMXrRIQIp';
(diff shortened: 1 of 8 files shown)


src/constants.js
```diff
@@ -0,0 +1,44 @@
+const constants = {
+  LOGIN_TOKEN_NAME: 'loginToken',
+
+  // AIRTABLE
+  BASE_ID: 'appFaOwKhMXrRIQIp',
+  ENDPOINT_URL: 'https://api.airtable.com',
+  GRID_VIEW: 'Grid view',
+  NUM_RECORDS: 1,
+
+  // TABLES
```
aivantg: this is definitely a great idea in terms of organization! If this file ends up being big and unwieldy, we could split it up into multiple like have a general `constants` object but also have one dedicated to table names and stuff like that. just a thought for the future!

PR #43, opened 2020-01-06
[1] setup basic jest testing infrastructure

dfangshuo:
### Related PRs
https://github.com/calblueprint/peoplepower-web/pull/44

### Screenshots
image

CC: @aivantg

### DIFF
--- a/src/__tests__/lib/request.spec.js
+++ b/src/__tests__/lib/request.spec.js
@@ -1,0 +1,15 @@
+import constants from '../../constants';
+import { getRecord } from '../../lib/request';
+
+const { PERSON_TABLE } = constants;
+
+const inputResultMapping = {};
+const TEST_PERSON_RECORD = 'recCxB46wmgbGEgi7';
+inputResultMapping[TEST_PERSON_RECORD] = 'TEST - DO NOT DELETE';
+
+describe('getRecord function', () => {
+  test('expect resolve', async () => {
+    const res = await getRecord(PERSON_TABLE, TEST_PERSON_RECORD);
+    expect(res.record.ID).toBe(inputResultMapping[TEST_PERSON_RECORD]);
+  });
+});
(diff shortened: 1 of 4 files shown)


src/__tests__/lib/request.spec.js
```diff
@@ -0,0 +1,15 @@
+import constants from '../../constants';
+import { getRecord } from '../../lib/request';
+
+const { PERSON_TABLE } = constants;
+
+const inputResultMapping = {};
+const TEST_PERSON_RECORD = 'recCxB46wmgbGEgi7';
+inputResultMapping[TEST_PERSON_RECORD] = 'TEST - DO NOT DELETE';
```
aivantg: I don't know much about the convention for writing Jest Tests, but the purpose of the `inputResultMapping` object isn't fully clear to me just from this. I'm assuming we use an object so it's easy to test multiple at a time in the future? 

Regardless, I think it would be worth having some minimal documentation for these tests that includes a link to maybe where someone can learn more about how to write them? Just so we have a shared standard for how they look. (might be overkill, you have the most perspective here, but just a thought!)

aivantg: Will we always have to have this Test record? I wonder if in the future we'll want tests that just validate that the structure of the data in the production database is valid. Not sure how to deal with that.., but good for development regardless!

dfangshuo: The structure of the test has changed a little bit with the next PR. Heard you on all of these points and I'll maybe improve the naming there: https://github.com/calblueprint/peoplepower-web/pull/44

dfangshuo: will improve naming in https://github.com/calblueprint/peoplepower-web/pull/44

Currently, there is a completely separate table in Airtable that this test writes to/reads from called Test(Development), and I'm not sure if this is something we want in production. The reason I went with it this time is because this test is meant to test the foundational functions that the rest of the backend is built on top off, and so I chose to err on the side of robustness by actually making the API call and verifying the data (as opposed to something like stringent, mocking up data etc.)

PR #46, opened 2020-01-09
[3] Fix bug: correct navbar state management 

dfangshuo:
## Fix bug: correct navbar state management 

Describe your feature here

### Related PRs

1. https://github.com/calblueprint/peoplepower-web/pull/43
2. https://github.com/calblueprint/peoplepower-web/pull/44

### Screenshots

CC: @aivantg @dfangshuo

### DIFF
--- a/src/components/NavBar.js
+++ b/src/components/NavBar.js
@@ -1,87 +1,31 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import '../styles/NavBar.css';
-import { getPersonById } from '../lib/request';
-import { getLoggedInUserId } from '../lib/auth';
-import applyCredentials from '../lib/credentials';
+import {
+  isAdmin,
+  isSubscriberOwner,
+  isGeneralOwner,
+  isSignedIn
+} from '../lib/credentials';
 import Logo from '../assets/PPSC-logo.png';
 
-export default class NavBar extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      id: '',
-      name: '',
-      credentials: ''
-    };
-  }
-
-  componentDidMount() {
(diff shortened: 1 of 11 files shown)


.eslintrc.json
```diff
@@ -8,6 +8,29 @@
         "extensions": [".js", ".jsx"]
       }
     ],
+    "no-restricted-imports": [
```
aivantg: I'd imagine with all these new PRs, there's a lot of shared history between the branches. Ideally to keep it simple as possible, just have the PR include the bugfix but it's not a big deal. I'll just review the relevant stuff for this PR

src/components/NavBar.js
```diff
@@ -1,92 +1,78 @@
...
     return (
       <div className="navBar">
+        <a href="/">
+          <img
+            className="logo"
+            src={Logo}
+            alt="People Power Solar Cooperative Logo"
+          />
+        </a>
+        <nav>
```
aivantg: We'll actually want a third version of the navbar for if they are currently going through onboarding it doesn't show "Sign in". But that's not crucial

src/lib/credentials.js
```diff
@@ -1,43 +1,56 @@
...
+  }
+
+  if (ownerTypes.includes(GENERAL_OWNER)) {
+    credentials += 'G';
+  }
+
+  return credentials;
+}
+
+function isSignedIn(credentials) {
```
aivantg: fantastic abstraction barrier. I didn't think of this when Nick and I first wrote the credential code but this is perfect!

.eslintrc.json
dfangshuo: good idea, will do

PR #47, opened 2020-01-09
Cleaned up CSS by factoring out commonly used components 

dfangshuo:
## Cleaned up CSS by factoring out commonly used components 

Main idea is to factor out common styles into a `main.css` file e.g. pink rounded buttons, square gray buttons etc.

commonly used styles in `main.css`
cc-ing @ethanlee16 on this one cause im not 100% sure what the best practice is here

### Screenshots

CC: @aivantg @dfangshuo


src/styles/main.css
```diff
@@ -0,0 +1,63 @@
...
+  --pp-white: #edf1f5;
+
+  --pp-light-gray: #f5f5f5;
+  --pp-bg-white: #efefef;
+  --pp-tag-gray: #d8d8d8;
+  --pp-link-gray: #8f8f8f;
+}
+
+.pp-pink-rounded-button {
+  -webkit-transition-duration: 0.4s; /* Safari */
```
ethanlee16: You shouldn't need any vendor-specific prefixes on your properties, since [CRA ships with Autoprefixer](https://create-react-app.dev/docs/post-processing-css/).

src/screens/onboarding/Onboarding.js
```diff
@@ -85,7 +85,7 @@ class Onboarding extends React.Component {
         billingState: '',
         billingZipcode: ''
       },
-      step: 1
+      step: 2
```
ethanlee16: Is this change intentional? It doesn't seem related to this PR, or I can't find anything on why this is included.

dfangshuo:
Cleaned up CSS files (including deleting some unused files) and switched things over to use kebab case consistently. Might leave switching to some class name convention in a future PR because i trie doing it and realized the changes would be too massive

PR #48, opened 2020-01-10
Auto-Generated Request.js and Schema.js

aivantg:
## Auto-Generated Request.js and Schema.js

This PR integrates the auto-generated `request.js` and `schema.js` files from  https://github.com/aivantg/airtable-schema and adjusts the codebase to work with the changes.

It also adds the `schema.json` file to the repo, which keeps track of the schema used to generate 
`request.js` and `schema.js`. In future runs of the generator, this `schema.json` should be used! 

Details on how to use the generator can be found at: https://github.com/aivantg/airtable-schema

## Screenshots

image
Auto-generated `request.js` and `schema.js`

image
Sample customized `schema.json` and resulting functions in `request.js`

CC: @dfangshuo


src/lib/credentials.js
```diff
@@ -22,7 +23,7 @@ export default function applyCredentials(userID) {
         return getOwnerById(ownerID);
       })
       .then(ownerRecord => {
-        const ownerTypes = ownerRecord[OWNER_TYPE_FIELD];
+        const ownerTypes = ownerRecord[Columns.Owner.OwnerType];
```
dfangshuo: Did the linter allow this :o It seems like the linter prefers object destructuring notation a la line 6 and idk if we want to do that here to for consistency's but if it's chill with the linter it's chill with me tbh

src/lib/schema.js
```diff
@@ -0,0 +1,247 @@
+/*
```
dfangshuo: this is really cool 😄

src/lib/credentials.js
aivantg: I'm not sure but I highly prefer the schema constants to be written this way for readability honestly. Maybe we can revisit the linter's obsession with destructuring at some point haha

PR #52, opened 2020-01-11
Async/Await + this.bind Style Updates 

aivantg:
I replaced almost all uses of `.then` promise chains with `async/await` code for consistencies' sake. I also documented when to use async/await vs regular promises in notion. General rule: default to async/await unless it logically flows better using `.then`. 

I also removed all calls to `this.<function> = this.<function>.bind(this);` and replaced them with arrow functions. 

These are both opinions for clean code, just for consistency. Feel free to comment if you disagree in spots! Tried to document these decisions a little here: https://www.notion.so/calblueprint/Style-Guide-05a67ecd1d0a41fdad3671e8aec0074e

Note: a lot of our code in downloading user info is suuuuper messy. Some would be fixed in the process of adding Redux or cleaning up the onboarding code, but part of me wonders if some should be left for developers to look over for themselves so they learn from earlier mistakes too. That said, there will be plenty more mistakes to write and fix next semester haha. 

CC: @aivantg @dfangshuo


src/lib/paypal.js
```diff
@@ -182,6 +182,13 @@ const recordBillPaymentSuccess = async (details, data, bill) => {
...
       it could cause problems
     */
+
+  try {
+    const paymentId = await createPaymentRecord(record);
+    console.log(paymentId);
+  } catch (err) {
+    console.error(err);
+  }
   createPaymentRecord(record)
```
dfangshuo: hmm should this createPaymentRecord call be deleted then since you're putting createPaymentRecord in a try catch?

src/screens/onboarding/ProjectGroups.js
```diff
@@ -20,29 +20,26 @@ class ProjectGroups extends React.Component {
...
-          default: record['Is Default?']
-        })
-      );
-      const selectableGroups = projectGroups.filter(
-        group => group.public && !group.default
-      );
-      const defaultGroup = projectGroups.find(group => group.default);
-      this.setState({ groups: selectableGroups, defaultGroup });
-    });
+  async componentDidMount() {
```
dfangshuo: these seem like a good candidate to put into a utils function instead of stuffing this logic here - we would be able to put a test around that too

src/screens/general/Community.js
```diff
@@ -36,39 +35,29 @@ export default class Community extends React.Component {
...
-        });
-        return getAnnouncementsByProjectGroup(projectGroup[0]);
-      })
-      .then(payload => {
-        this.setState({
-          cards: payload,
-          isLoading: false
-        });
-      });
+    const personRecord = await getPersonById(id);
```
dfangshuo: This chunk seems like a good candidate to put into a utils function

src/screens/onboarding/Onboarding.js
```diff
@@ -88,71 +88,65 @@ class Onboarding extends React.Component {
...
-          projectGroup: person['Project Group'][0]
-        });
-        const { Owner: owner } = person;
-        return getOwnerById(owner);
-      })
-      .then(owner => {
-        if (step > 3) {
-          const numShares = owner['Number of Shares'];
+    this.setState({
+      step: person['Onboarding Step'], // TODO: move this object transformation out of our visual components
```
dfangshuo: We should probably use fields for these strings from the airtable schema file you generated

src/screens/onboarding/Onboarding.js
```diff
@@ -88,71 +88,65 @@ class Onboarding extends React.Component {
...
-            this.setState({
-              bylaw1: true,
-              bylaw2: true
-            });
-          }
-        }
+      this.setState({
+        projectGroup: owner['Project Group'][0],
+        numShares: numShares || 1,
+        dividends: owner['Receiving Dividends?']
```
dfangshuo: ditto the above comment about using fields from the airtable schema

src/screens/general/UserProfilePage.js
```diff
@@ -39,89 +39,60 @@ export default class UserProfilePage extends React.Component {
...
-    this.handleSubmit = this.handleSubmit.bind(this);
   }
 
-  componentDidMount() {
+  async componentDidMount() {
     // id taken from URL. React Router's useParams() threw an "invalid hook" error.
     const { match } = this.props;
     const { id } = match.params;
+    const personRecord = await getPersonById(id);
+    const {
```
dfangshuo: another good candidate to put into a utils function imo

dfangshuo: also we should probably use the fields in the airtable schema file here as well?

src/screens/general/GeneralOwnerDashboard.js
```diff
@@ -37,83 +37,56 @@ export default class GeneralOwnerDashboard extends React.Component {
...
-    let email;
-    let phoneNumber;
-    let name;
-    let owner;
-    let city;
-    let street;
-    let zipCode;
-    let state;
+    const personRecord = await getPersonById(personId);
+    const {
```
dfangshuo: a good candidate to put into a utils function we can test.

src/lib/paypal.js
aivantg: good catch!

aivantg:
> 1. (i probably should have flagged this earlier too tbh) we have a number of places in our components where we have a lot of logic that would be great candidates for helper functions living in a utils file (that we can test). I left a couple of these comments at various places.


Yeah, agreed, I decided to solely transform the promises into async/await without any code editing. There's a couple of things that are perfect for a utils file and a couple of things that will be made 10x simpler with redux (like literally 25 lines to 1 or 0) so I figured it'll be easier to solve both at once! 

> 2. we use a lot of strings that we can probably replace with constants from the airtable schema file you generated

And yeah, we should definitely use the schema constants more, but I hesitated to fix them all in this PR because I realize we actually have to fix them in a LOT of places. Like every time we translate a record object to variables we basically reference specific column names. Sometime in the next few days I wanna see if I can clean up the onboarding files and extract some more utility functions and then i'll experiment with simplyfing our record => variables transformation! 

> 1 point of clarification for the airtable schema field -> each time we modify the schema we need to manuallly run and regenerate the schema file?

Yes, the steps to regenerate would look like: 
1) copy the schema.json file from the project into the output folder of the schema-generator repo. 
2) run `npm start` in the other repo
3) copy back all 3 files back into the project

You have to do step 1 because that way it preserves any custom lookup fields that you might add to the schema.json file

PR #54, opened 2020-01-12
Working Subscriber Owner Dashboard + Correct Navbar Mistake

dfangshuo:
## Working Subscriber Owner Dashboard + Correct Navbar Mistake

The previous version of the subscriber owner dashboard (on Master) uses a frontend that is built very hackily. This PR corrects that by using the package `react-table-v6`.

Additionally, this dashboard also shows Payments (like Iris intended in the Figma).

@aivantg The 1 thing thing I wasn't really able to figure out was how to get rid of the gridlines from the react-table. Like if you look at Iris' figma designs, she doesn't have those vertical and horizontal grid-lines across rows/columns, but i didn't manage to find a way to implement that. If you are able to find a fix for that thatt would be 💯 

Additionally, also corrected an earlier error where the navbar wasn't updated correctly. Included the user's name in the loginToken so as to make the rendering/updating of the navbar more convenient

[//]: # (Describe your feature here)

### Related PRs

[//]: # (Optional - any related PRs you're waiting on, or PRs that will conflict, etc)

### Migrations

[//]: # (Optional - if you added anything to the database through migration[s])

[//]: # (If you made changes/additions to `airtable.js` or `request.js`,) 
[//]: # (make sure you update/add [at least 1] corresponding unit tests in)
[//]: # (`airtable.spec.js` or `request.spec.js`)

### Screenshots

image
image

CC: @aivantg @dfangshuo


src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -1,19 +1,37 @@
...
+
+// expected format of date: YYYY-MM-DD
+function dateToDateString(date) {
+  const dateArr = date.split('-');
+  return `${dateToWord[parseInt(dateArr[1], 10)].substring(0, 3)} ${
+    dateArr[2]
+  }, ${dateArr[0]}`;
+}
 
+function generateAllBillsHeader(headerText) {
```
aivantg: I think a the "conventional" way to write a function like this (besides extracting it to its own component, which would be great but feels like overkill) is to call it `renderAllBillsHeader` and make it a method inside the component (instead of outside). 

I'm saying this solely from how i've seen it in other codebases haha, not for any particular logical reason. Regardless, I think it would still be good to put it inside the component since it's only used here!

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -1,19 +1,37 @@
...
+  return () => (
+    <div className="subscriber-all-bills-row subscriber-all-bills-header">
+      {headerText}
+    </div>
+  );
+}
+
+export default class SubscriberOwnerDashboardAllBillsView extends React.Component {
```
aivantg: man this name is a mouthful. We can revisit component naming convention one day as a team if it becomes partiucalrly annoying

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -23,8 +41,54 @@ export default class SubscriberOwnerDashboardAllBillsView extends React.Componen
   }
 
   render() {
-    const { bills } = this.state;
-    const { callback } = this.props;
+    const { callback, transactions } = this.props;
+    const data = transactions.map(transaction => {
```
aivantg: this might be another good place to extract some rendering out to a component or another function. Like maybe a `renderTransaction` function, or a `renderOnlineTransaction` and a `renderBillTransaction` to help keep our main render function clean! 

Whether we want to go as far as make some of these things their own component is a design decision. Some say you always should and components should all be like less than 100 lines, but idk sometimes that's overkill (like with the `AllBillsHeader` thing from above). Up to you!

aivantg: but yeah that would make the line above look something like: 
`const data = transactions.map(t => (<Transaction transaction=t />))`
or 
`const data = transactions.map(this.renderTransaction)`
or 
`const data = transactions.map(t => (t.Type === ONLINE_PAYMENT_TYPE ? this.renderPaymentTransaction(t) : this.renderBillTranasaction(t)))`

I'm sooorta leaning towards making it its own component but I don't know this part of the codebase well enough

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -23,8 +41,54 @@ export default class SubscriberOwnerDashboardAllBillsView extends React.Componen
...
+      if (transaction.Type === ONLINE_PAYMENT_TYPE) {
+        return {
+          balance: '$0.00',
+          statementDate: dateToDateString(transaction['Transaction Date']),
+          description: transaction.Type,
+          status: formatStatus(transaction.Status),
+          payment: `$${centsToDollars(transaction.Amount)}`
```
aivantg: once again, any reason the attributes of the transaction object are capitalized? I feel like that goes against convention and how we've been doing it in the rest of our code

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardMainView.js
```diff
@@ -1,13 +1,37 @@
...
-import PanelBillHeader from './PanelBillHeader';
-import PanelBillRow from './PanelBillRow';
+
+import constants from '../../../constants';
+
+const { ONLINE_PAYMENT_TYPE } = constants;
 
 const clientId = process.env.REACT_APP_PAYPAL_CLIENT_ID;
 
+function generateBillDisplayHeader(headerText) {
```
aivantg: if this function is used in two different files, that's one more point for making a component out of it, but I still think it's overkill lol. I'd recommend the same changes as i did for the `AllBillsView` component though

aivantg: Same for the date functions below!

aivantg:
I think I managed to remove gridlines by overriding the `getTdProps` prop for the <ReactTable /> Element. aka adding: 
```
getTdProps={() => ({
    style: { border: `none` }
})}
```
as a prop to the ReactTable component. Looking at documentation [here](https://www.npmjs.com/package/react-table-6#styles), seems like you can override a bunch of their native table elements and you can pass in both a style and a class name prop (so we can still keep our style code in css). 

I think the problem with what you had currently miiiight have been that the class you applied applies to the entire ReactTable container and not to the individual `td` elements. Idk, give it a try to see if it accomplishes what you're looking for!

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
dfangshuo: This used to be a shorter name but I think we agreed on this naming convention after 1 PR last semester. But yeap ill keep an eye on this

src/screens/general/AdminDashboard.js
```diff
@@ -47,6 +47,9 @@ export default class AdminDashboard extends React.Component {
         owners: owners.filter(owner => owner.Person[0] !== personId),
         isReady: true
       });
+      const { updateState } = this.props;
+      const name = getLoggedInUserName();
```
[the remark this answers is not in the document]
dfangshuo: essentially - yes! not doing so leads to some inconsistency with how the navbar is displayed - e.g. if you logged in, close the tab, and then go to the AdminDashboard directly from a new tab, the navbar will not display the right state if updateState isn't called.

Agree with your comment that this is annoying, am looking forward to redux and/or to other ways of structuring this

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboard.js
```diff
@@ -27,27 +27,21 @@ export default class SubscriberOwnerDashboard extends React.Component {
...
-    this.setState(prevState => {
-      if (areDiffBills(prevState.transactions, transactions)) {
-        return { transactions, isReady: true };
+      const { updateState } = this.props;
+      const name = getLoggedInUserName();
+      updateState(personId, name);
+      const transactions = await getSubscriberBills(personId);
+      if (transactions) {
+        this.setState(prevState => {
+          if (areDiffBills(prevState.transactions, transactions)) {
```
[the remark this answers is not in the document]
dfangshuo: Ahhh yes this is just to prevent an infinite loop. Without checking if the bills are different the function loops forever and the screen freezes

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -1,19 +1,37 @@
...
-    super(props);
-    const { transactions } = this.props;
-    this.state = {
-      bills: transactions
-    };
-  }
+const { ONLINE_PAYMENT_TYPE } = constants;
+
+function dateToFullMonth(date) {
+  return dateToWord[parseInt(date.split('-')[1], 10)];
```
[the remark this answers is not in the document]
dfangshuo: That's a great idea, I think the way I'm going to do this is move all of the date utilities into a file like dateUtils, and then make the change to a library in a future PR

src/lib/subscriberUtils.js
```diff
@@ -1,7 +1,22 @@
...
+const { BILL_TYPE, ONLINE_PAYMENT_TYPE } = constants;
+
+const formatStatus = status => {
+  return (([firstLetter, ...rest]) =>
+    [firstLetter.toLocaleUpperCase(), ...rest].join(''))(status.toLowerCase());
+};
+
+const convertDateTimeToDate = dateTimeString => {
+  if (dateTimeString) {
+    return dateTimeString.slice(0, 10);
```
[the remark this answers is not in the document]
dfangshuo: mm good point, I think this is only meant to handle the date format returned from the paypal api, and ive renamed it as such.

This is not the most sustainable solution, but I've read your other comments and i intend to make the migration to moment.js in a future PR.

src/lib/subscriberUtils.js
```diff
@@ -77,24 +106,45 @@ const getSubscriberBills = async (loggedInUserId, callback) => {
...
     }
 
-    callback(bills);
+    if (paymentObjects) {
+      paymentObjects.forEach(paymentObject => {
+        transactions.push({
+          'Transaction Date': convertDateTimeToDate(
+            paymentObject['Payment Create Time']
+          ),
+          Type: ONLINE_PAYMENT_TYPE, // Type is a local variable inserted to distinguish between bill payments and online payments
```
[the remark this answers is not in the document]
dfangshuo: Hmm I tried making the change but I think this has got to do with the old reason we're recording payments in airtable. It's a little more complex than I first thought and I'll address this in a future PR

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -1,19 +1,37 @@
...
+}
+
+// expected format of date: YYYY-MM-DD
+function dateToDateString(date) {
```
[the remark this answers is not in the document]
dfangshuo: will address in future PR

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
dfangshuo: will address in future PR

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
dfangshuo: Mm, this is more of a ESLint thing because this function doesn't depend on any state or props, so the convention (according to the linter) is to put it outside the component

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
dfangshuo: didn't adhere to this exactly, but I pulled some stuff out into its own helper functions/components. 

TLDR:

I made the table displaying all bills a component of it's own, since it's fairly complex, and c created helper functions adhering to this pattern you suggested:

`const data = transactions.map(t => (t.Type === ONLINE_PAYMENT_TYPE ? this.renderPaymentTransaction(t) : this.renderBillTranasaction(t)))`

dfangshuo:
1. Agree that keep calling `updateState` is a little annoying, looking forward to Redux :-)
2. extracted the date utilities into a separate helper file, will do the migration to moment.js or some other library in a future PR
3. Will also address the Capitalization isssue in a future PR

Any tips or pointers to any resource I can use to get familiarized with redux?

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
aivantg: lol i vote to remove this eslint thing. I think part of the reason is because eslint is pushing you to either 1) extract to a component or 2) extract to a utils file. Which I understand but is kinda annoying sometimes for simple things

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboard.js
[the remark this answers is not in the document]
aivantg: huh...weird ¯\_(ツ)_/¯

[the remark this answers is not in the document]
dfangshuo: HAHA you all talked about it during 1 of the modules for the React Decal (I can't remember which though), about how setState triggers some change or something and how you have to copmare change to prevState otherwise it will infinitely recursively trigger changes

PR #57, opened 2020-01-13
Update Paypal Payment flow

dfangshuo:
## Update Paypal Payment flow

[//]: # (Describe your feature here)
The original implementation of Paypal payments only allows you to pay your latest bill. This means that if you have multiple unpaid bills, you can still only pay for your latest bill, and must repeatedly pay to clear multiple bills.

This PR changes that by making your balance be the total outstanding balance of multiple bills, meaning if you have multiple bills, you can pay for them all at once.

The limitation of the current implementation is a step in the right direction but as it is right now, you have to pay for the full amount owed. This means that if you have 3 unowed bills totaling to $281.21 let's say, you have to make a payment of $281.21. There are no partial payments supported. *The plan is to change this in a future PR.*

Additionally, a payment that pays multiple bills records a billl payment for each bill. Aka if you pay $281.21 bills for 2 bills that aree $200 and $82.21 respectivelly, the payment is logged into the Payment table as 2 payments of $200 and $82.21, with the same paypal metadata (because these were paid at the same time). *This might also be changed in the future*

This PR also updates a bill's status to complete once a bill is successfully paid (this was not implemeented before).

*also updated the styling of the paypal button and resolved the struct capitalization issue from https://github.com/calblueprint/peoplepower-web/pull/54

### Related PRs

[//]: # (Optional - any related PRs you're waiting on, or PRs that will conflict, etc)

### Migrations

[//]: # (Optional - if you added anything to the database through migration[s])

[//]: # (If you made changes/additions to `airtable.js` or `request.js`,) 
[//]: # (make sure you update/add [at least 1] corresponding unit tests in)
[//]: # (`airtable.spec.js` or `request.spec.js`)

### Screenshots

CC: @aivantg @dfangshuo

### DIFF
--- a/src/lib/paypal.js
+++ b/src/lib/paypal.js
@@ -1,10 +1,15 @@
 import { createPayment, updateOwner, updateSubscriberBill } from './request';
 import constants from '../constants';
 
-const { BILL_PAYMENT_TYPE, BUY_SHARES_TYPE } = constants;
+const {
+  BILL_PAYMENT_TYPE,
+  BUY_SHARES_TYPE,
+  COMPLETED_STATUS,
+  PENDING_STATUS
+} = constants;
 
-const createPaymentRecord = async record => {
-  /*
+/*
+  Schema of Payment Table
       {
         "fields": {
           "Owner": [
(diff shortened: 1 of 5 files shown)


src/lib/paypal.js
```diff
@@ -19,8 +24,15 @@ const createPaymentRecord = async record => {
...
-  return id;
+
+const dollarsToCents = dollars => {
+  return parseFloat(dollars, 10) * 100;
+};
+
+const getTotalBalanceFromBills = pendingBills => {
+  return pendingBills
+    .map(pendingBill => pendingBill.Balance)
+    .reduce((a, b) => a + b, 0);
```
aivantg: love myself some JS array functions 😍 

Something we should double check with Grayson about is how bills work. Is each month independent or is it more like a "statement balance" that aggregates previous amount owed and new amount due.

PR #59, opened 2020-01-14
[1] Set up Redux to Cache User Data

aivantg:
### What Happened in this PR

**I installed redux into the project!**

This has made so much of our code blissfully more simple!

Full info on how it works is here: https://www.notion.so/calblueprint/PP-Power-Redux-0735d9d4783f4100b5d79e180e96c61b

Our store has two main sections: one for user data and one for community
- The user data slice of the store stores: the person record, owner record, userLogin record, projectGroup record, solar project records, credential info, and authentication info
- The community slice of the store stores the announcement array. 

Every main screen has been refactored to get the information it needs from the redux store instead of asking airtable. 

The redux store pulls from airtable 1) when the user logs in 2) when the app loads (if the user was already logged in), 3) when the user updates their profile (currently not working due to a bug) 

**I rearranged some folders**

The lib folder was getting very large so I organized some files within it into sub-folders. I also combined the images and assets folder

### What doesn't work
While I've tested most of the main app's functionality, It's not without bugs. For example, for some reason you can't select a project group in onboarding anymore. Onboarding is sort of a complicated question under redux so I think we should try and figure that out in a separate PR.

### What comes next? 
These things can 100% be tackled during the semester as well as before, just some thoughts as I was refactoring code

- Bring routing into Redux (I plan on adding another PR for this soon)
- Figure out the best way to reduxify onboarding. 
- Refactor Airtable tables to simplify development
- Transform airtable records on read and write at a lower level (in `airtable.js`) instead of in every single component

### How to review this PR

I tried my best to limit the refactoring to what was necessary and making TODO's where I saw problems to fix in the future. But I'm sorry this is a MASSIVE PR. Please ask any and all questions you have and I'm happy to go through it and explain everything over video chat. I genuinely think all of our devs should read through and understand the changes made in this PR since so much of our codebase has been simplified through redux.

Order of which to review files: 
1. Check out the files in `lib/redux/`. The two slice files define the structure of their part of the store as well as the actions available. This is the core of our redux code
2. Check out `userDataUtils.js` for the main data refresh function
3. Check out `Login.js` and `App.js` to see how they "request" a fresh copy of the data when a user logs in 
4. Take a look at a simple component like `Navbar.js` to see how a component can read data from the store using `mapStateToProps`
5. Go through the rest of the components and files to see how they too can be simplified by reading data from redux instead of requesting airtable

### What do I need to install to test it? 

Redux DevTools is a lifesaver while debugging redux: https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en

CC: @aivantg @dfangshuo


aivantg:
@ethanlee16

FranCogMents:
😲 this is an amazing PR , ++ @aivantg  on using the redux tool kit!

I was wondering if you  + @dfangshuo could explain the motivation behind adding redux to the project? Specifically what problems does redux help address for this project? It sounds like one of the use cases is is to centralize data access from airtable so that components no longer need to interact with airtable directly. I was wondering what helped you all decide between using redux vs React Context in regards to your use cases?

aivantg:
> 😲 this is an amazing PR , ++ @aivantg on using the redux tool kit!
> 
> I was wondering if you + @dfangshuo could explain the motivation behind adding redux to the project? Specifically what problems does redux help address for this project? It sounds like one of the use cases is is to centralize data access from airtable so that components no longer need to interact with airtable directly. I was wondering what helped you all decide between using redux vs React Context in regards to your use cases?

Thanks! The main use case right now is simply using it as a local cache for all of our airtable data. We fetch it once and then our components get the data from the store instead of calling airtable. There are definitely a number of ways to solve this, and I looked into alternatives like React Context. I think I settled on Redux because 1) I'm already pretty familiar with it 2) I prefer the syntax and 3) after I read up on Redux Toolkit my fear that Redux was "too much" kind of went away since it simplifies a lot of things. So with all that, i decided to favor redux!  Not super strong reasons, but after a bunch of research I decided to just make the jump :)

dfangshuo:
am going  through it right now, but something i'm a little curious about  at this point is why did you decide to split the slices the way that you did (userData and community)? I went through the Notion and saw you mentioning that it's good to separate out slices by features, but why/how did you decide that userData was 1 feature and community was another? 

Not really a problem or anything, really just for my own understanding of how you structured this @aivantg

aivantg:
> am going through it right now, but something i'm a little curious about at this point is why did you decide to split the slices the way that you did (userData and community)? I went through the Notion and saw you mentioning that it's good to separate out slices by features, but why/how did you decide that userData was 1 feature and community was another?
> 
> Not really a problem or anything, really just for my own understanding of how you structured this @aivantg

No _real_ reason honestly. I actually wondered after I finished whether it was worth splitting them in the first place. But in the moment, the reason that I did it was because it felt like a slight separation of concerns. All the data stored under the `userData` slice was data directly associated with the user: the 3 user records, their project group, etc. The `community` slice felt more like "all the community records" with a small filter. It definitely could be grouped with the user under the same logic as project group, but ¯\_(ツ)_/¯ yeah just a small design decision i made on the spot.

Since we're using redux as sort of a very simple airtable cache right now, it doesn't really matter very much. If we end up using redux as more than that, we might want to rethink our organization a little bit more.

src/screens/onboarding/Complete.js
```diff
@@ -1,14 +1,24 @@
...
   }
 
-  dashboardButton = () => {
-    const { history, toggleNavbar } = this.props;
+  dashboardButton = async () => {
+    const { history, toggleNavbar, values } = this.props;
     toggleNavbar();
+
+    // TODO: Replace with proper airlock authentication
+    store.dispatch(authenticate('temp_token'));
```
dfangshuo: so every user who authenticates would have the key 'temp_token'?

src/screens/onboarding/ContactInfo.js
```diff
@@ -116,7 +117,8 @@ class ContactInfo extends React.Component {
           const { handleRecordCreation } = this.props;
           handleRecordCreation({
             createdOwnerId,
-            createdPersonId
+            createdPersonId,
+            createdUserLoginId
```
dfangshuo: so for we're now also setting `createdUserLoginId` with `handleRecordCreation` now?

src/screens/general/UserProfilePage.js
```diff
@@ -145,6 +148,12 @@ export default class UserProfilePage extends React.Component {
         zipcode: updateZip
       });
     }
+
+    // get latest copy of user login record
+    const userLogin = await getUserLoginById(userLoginID);
```
dfangshuo: hmm is there a reason why we need to getUserLogin here and then update redux instead of getting it from redux?

src/lib/credentials.js
```diff
@@ -1,25 +1,23 @@
...
 
-  if (userId == null) {
+  // TODO: Will this ever be the case? Is this check necessary
+  if (owner == null) {
     return credentials;
   }
 
-  const ownerId = await getOwnerFromPerson(userId);
-  const isAdminPayload = await getAdminTable(ownerId);
+  const isAdminPayload = await getAdminTable(owner);
```
dfangshuo: hmm a little unclear what `getAdminTable` is supposed to do (repeat comment as the same comment I left at `getAdminTable`)

src/lib/subscriberUtils.js
```diff
@@ -30,23 +29,13 @@ const centsToDollars = cents => {
...
-    const ownerId = await getOwnerIdFromPersonId(loggedInUserId);
-    const { billIds, paymentIds } = await getBillsAndPaymentsFromOwnerId(
-      ownerId
-    );
+    const billIds = owner[Columns.Owner.SubscriberBill];
+    const paymentIds = owner[Columns.Owner.Payment];
 
     if (!billIds && !paymentIds) {
-      return { transactions: [], totalBalance: 0 };
+      return { transactions: [], pendingBills: [], totalBalance: 0 };
```
dfangshuo: should remove `totalBalance` here to conform with what the rest of the function is returning

src/lib/redux/store.js
```diff
@@ -0,0 +1,29 @@
...
+  key: 'root',
+  storage
+};
+
+const persistedReducer = persistReducer(persistConfig, reducers);
+
+// Set up store with the appropriate reducers
+const store = configureStore({
+  reducer: persistedReducer,
+  middleware: [thunk]
```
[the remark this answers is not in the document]
aivantg: Haha good question. I was reading that redux-thunk was default included when you used Redux Toolkit or something, so when I was installing this I think I left in the thunk middleware.

But a genuine use case for redux-thunk that I was thinking about was combining our `fetchAllData` redux actions into a single action (not two separate ones for loading and saving). I think redux-thunk is the ideal way to do async in redux. Was gonna save that for a future PR though. I can remove for now though!

src/screens/onboarding/Complete.js
aivantg: yeah for now. we don't actually use this key yet either. but eventually this will need to be replaced with an Airlock Auth Token so I put it in our store so we're prepared to interact with it

src/screens/onboarding/ContactInfo.js
aivantg: yeah, i stand by NONE of the code in onboarding right now haha. I tried to stick to solely refactoring what was necessary to make it work like it used to rather than cleaning it up. That's to come in a future PR

src/screens/general/UserProfilePage.js
aivantg: I designed the `fetchUserData` function to take in a user login record because in `Login.js`, it has to fetch the relevant user login record anyway (when it tries to find one based on email and password). This way it doesn't need to re-fetch it. 

I guess that means though that the user login record would never get updated....which I _think_ is fine but...maybe should note that.

src/lib/credentials.js
aivantg: lol once again, i have no idea either. i refactored it to work with redux but left refactoring the admin code to another PR (also.....i _think_ you wrote this function 😂

src/lib/subscriberUtils.js
aivantg: ditto to above. refactor later

src/lib/userDataUtils.js
```diff
@@ -0,0 +1,72 @@
...
+  store.dispatch(fetchAnnouncements());
+
+  // Fetch all the data
+  const ownerId = userLogin[Columns.UserLogin.Owner];
+  const personId = userLogin[Columns.UserLogin.Person];
+
+  const owner = await getOwnerById(ownerId);
+  const person = await getPersonById(personId);
+
+  const projectGroupId = owner[Columns.Owner.ProjectGroup];
```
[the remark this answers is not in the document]
aivantg: We don't have the owner record at that point. We only have the user login record :/

aivantg:
> ++ good stuff man, also really love the folder structure refactoring, makes things so munch clearer now 🌟
> 
> 1. what are your thoughts on moving the `redux` and `airtable` folderss out of `lib` so `src/redux` and `src/airtable`?

Up to you! Personally I like keeping the root folder as clean as possible, but it's 100% a personal thought. I'm down to move it outside too 

> 2. This was a huge PR I think i got most things but just in case I might have missed some stuff might take a look again after we discussed some of the comments I left

👍 thanks for reviewing!!

aivantg:
And yeah, @dfangshuo  I tried to NOT refactor any code that i felt was unclean and stuck with JUST making it work with Redux. I think we could use 2-3 more PR's refactoring 1) Admin Dashboard, 2) Subscriber stuff maybe?, 3) Onboarding, and 4) Anything else. But wanted to get these 3 in first :)

src/screens/general/UserProfilePage.js
dfangshuo: hmm wait hang on which `fetchUserData` function r u referring to

src/lib/subscriberUtils.js
dfangshuo: hmm.. agree with above, but in this case, i still think the right course of action would be to remove  `totalBalance` (instead of adding `pendingBills` and not removing `totalBalance`), because `totalBalance` is a relic from an old implementation, and leaving the codebase in a state where both `pendingBillls` and `totalBalance` is present might cause confusion in the future when people look at this (also `totalBalance` is not consumed anywhere)

src/screens/general/UserProfilePage.js
aivantg: oh whoops, the one in `UserDataUtils.js`

src/screens/general/SubscriberOwnerDashboard/SubscriberOwnerDashboardAllBillsView.js
```diff
@@ -3,15 +3,7 @@ import '../../../styles/SubscriberOwnerDashboard.css';
 import '../../../styles/SubscriberOwnerDashboardAllBillsView.css';
 import SubscriberOwnerDashboardAllBillsTable from './SubscriberOwnerDashboardAllBillsTable';
 
-export default class SubscriberOwnerDashboardAllBillsView extends React.Component {
-  componentDidMount() {
-    const { history, personId } = this.props;
-    if (!personId) {
-      // They shouldn't be able to access this screen
-      history.push('/');
```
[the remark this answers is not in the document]
aivantg: Yup! I have an idea to help do this at the router level, just haven't gotten around to it yet

src/screens/onboarding/Complete.js
```diff
@@ -1,14 +1,24 @@
...
 class Complete extends React.Component {
   constructor(props) {
     super(props);
     this.state = {};
   }
 
-  dashboardButton = () => {
-    const { history, toggleNavbar } = this.props;
+  dashboardButton = async () => {
+    const { history, toggleNavbar, values } = this.props;
```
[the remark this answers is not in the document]
aivantg: 👍

src/screens/onboarding/Onboarding.js
```diff
@@ -161,10 +161,15 @@ class Onboarding extends React.Component {
     this.setState({ step: step - 1 });
   };
 
-  handleRecordCreation = ({ createdOwnerId, createdPersonId }) => {
+  handleRecordCreation = ({
```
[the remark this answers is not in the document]
aivantg: all this function does is save the ID's. it's very redundant I'm sure lol but i opted to hold out to do this in a different PR

src/screens/general/UserProfilePage.js
dfangshuo: ah you mean `refreshUserData`?

dfangshuo: hmm ok i see what you mean, should probably add a comment documenting that the userLogin is passed in for that reason for clarity's sake?

src/lib/subscriberUtils.js
aivantg: fixed

src/screens/general/UserProfilePage.js
aivantg: sounds good!

PR #61, opened 2020-01-15
Installed Airtable Schema Generator as an NPM Package

aivantg:
## Installed Airtable Schema Generator as an NPM Package

Whenever you make a change to the Airtable API, all you have to run is `npm run generate-schema` and it will generate the new files. 

Heads up, you need to create a `.airtable-schema-generator.env` file as described in the [package readme](https://github.com/aivantg/airtable-schema-generator).

CC: @aivantg @dfangshuo


dfangshuo:
hmm, after i `checkout`-ed this branch and ran `yarn install`, it takes forever
image

wondering if you might have any idea why this is the case?

also:

1. could you maybe also link to your package README from the peoplepower-web README (maybe below the Quickstart sction)?
2. i'm assuming that `yarn generate-schema` also works, want to confirm if thatis the case?

aivantg:
> hmm, after i `checkout`-ed this branch and ran `yarn install`, it takes forever to run
> image
> 
> wondering if you might have any idea why this is the case?

It's because the airtable schema generator depends on `nightmare` which is the browser-automation npm package that lets the generator scrape the schema from airtable's website. Installing some of its dependencies take a bunch of time. We could look into shortening this but I don't think it's a big deal because 1) you should only have to install the library once and 2) it's a dev dependency, so it doesn't increase the bulk of our actual project size, it's just something you install locally so you can run the script

> also:
> 
> 1. could you maybe also link to your package README from the peoplepower-web README (maybe below the Quickstart sction)?

Yup good idea! will put that on my to-do list 

> 2. i'm assuming that `yarn generate-schema` also works, want to confirm if that is the case?

Just checked, it does!

aivantg:
Gonna hold off on merging this because I wanna make some more updates to the schema generator first

aivantg:
Made some updates to the schema generator to remove the use of the Constants file in `airtable.js` and to get rid of the `cleanTableName` function. Updated our repo to use the latest generator!

PR #64, opened 2020-01-22
[Easy] Enforce that new users must use a unique email

dfangshuo:
[//]: # (These comments meant for your reference, they are invisible and don't need to be deleted) 

## What's new in this PR?

[//]: # (Describe what's new in this PR)
Enforce that new users must use a unique email

### How to Review

[//]: # (The order in which to review files and what to expect when testing locally)

### Related PRs

[//]: # (Optional - related PRs you're waiting on/ PRs that will conflict, etc)

### Next steps

[//]: # (What doesn't work yet, what's NOT in this PR/has to be done )

[//]: # (If you made changes/additions to `airtable.js` make sure you update/add) 
[//]: # ([at least 1] corresponding unit tests in `airtable.spec.js`)

Used a non-unique email (aivantg@berkeley.edu) and verified that the account creation failed. Subsequently used a unique email (aivantgg@berkeley.edu) and verified that I can proceed to the next step

### Screenshots

CC: @dfangshuo


src/screens/onboarding/BasicInfo.js
```diff
@@ -19,6 +20,14 @@ class BasicInfo extends React.Component {
       password,
       altEmail
     } = values;
+
+    const isNewEmail = await isUniqueEmail(email);
+    if (!isNewEmail) {
+      // TODO: surface issue to user here
+      console.log('Email exists');
```
aivantg: Since we're in the onboarding part of the app, there actually is a way to surface this error to the user I think? Like instead of returning here you could add it to the error message array. I'd say while this is a quick fix, it's still worth integrating for the time being

aivantg: jk if i'm gonna revamp onboarding then it's fine :P

PR #65, opened 2020-01-26
[Easy] Integrated routing with redux

aivantg:

## What's new in this PR?

Using [Connected React Router](https://github.com/supasate/connected-react-router), integrated routing with Redux. Routing data is now stored in Redux so that you can access information about the current path from anywhere, including a component that isn't a normal screen (like the navbar!)

This is really just so that the navbar can read the current route and update itself based on it. Earlier, only components that were associated with a route got passed the `history` prop. 

### How to Review

1. `store.js` and `app.js` contains changes switching from normal react-router to connected react router (which adds a sort of "router" slice to the redux store with routing info)
2. `App.js`, `Navbar.js` and the onboarding components contain changes for removing the awkward `isNavbarVisible` and `toggleNavbar` props

Most edits for part 1 are just following the instructions in the `connected-react-router` readme step for step. 

All parts of the app should work exactly the same: special version of navbar should show during onboarding. 

### Next steps

- I want to clean up the navbar's conditional rendering. 
- I want to push up the unauthorized redirects to `App.js` so components no longer have to worry about whether a user is allowed to access that page. 

### Related PRs

#66 - follow-up PR addressing both of those next steps together and cleaning up credential stuff in general! 

### Screenshots

CC: @dfangshuo


src/lib/redux/store.js
```diff
@@ -19,9 +25,11 @@ const persistedReducer = persistReducer(persistConfig, reducers);
 
 // Set up store with the appropriate reducers
 const store = configureStore({
-  reducer: persistedReducer
+  reducer: persistedReducer,
+  middleware: [routerMiddleware(history)]
```
dfangshuo: hmm do you happen to know what routerMiddleware does and/or why it's in a list?, or is this part of following the `connected-react-router` setup?

src/components/NavBar.js
```diff
@@ -10,15 +10,16 @@ import {
 import Logo from '../assets/PPSC-logo.png';
 import '../styles/NavBar.css';
 
+const ONBOARDING_ROUTE = '/onboarding';
+
 class NavBar extends React.PureComponent {
   render() {
-    // TODO: move routing to redux and update navbar display based on route
-    const { person, credentials, isNavBarVisible } = this.props;
+    const { person, credentials, path } = this.props;
```
dfangshuo: hmm, so this path is what will be automatically supplied to `props` by redux when the store updates?

src/lib/redux/store.js
aivantg: I don't know _exactly_ how redux middleware works, but the concept of middleware is to basically let an outside library add its own configuration steps to your workflow.  But yes this is also part of the setup instructions.

src/components/NavBar.js
aivantg: yup, it's just another slice in the redux store that is automatically managed and updated by the connected-react-router package. The reason it can do this is because we added the middleware and all hehe . 

`path` is supplied to props only because i specified it in `mapStateToProps` at the bottom of the component

PR #66, opened 2020-01-26
Reorganized folders and cleaned up routing/authentication

aivantg:

## What's new in this PR?

This PR contains two major changes: 
- I reorganized all the folders and components to follow the suggested naming/organizational components I proposed earlier. (sorry should've been it's own PR but I did it amidst navigation changes)
- Updated `App.js` and `Navbar.js` to deal with authentication and re-routing in a clean, organized way. 

Now, when you define a route in `App.js` you specify who has access to this route by creating an `<AuthenticatedRoute />`. You can pass in a `credential` prop ('S', 'A', or 'G') to specify a specific type of user that can access the route. You can pass in a `noauth` prop to specify that the user _shouldn't_ be authenticated (like the onboarding route). Or you can pass in no props to just specify that the user should at least be logged in. 

`App.js` also now calculates which component should be shown at the `/` route so that it's nice and simple

`Navbar.js` has also been cleaned up to be a little more streamlined and less random if-statement logic

### How to Review

Take a look at the changes in `AuthenticatedRoute.js`, `App.js`, and `NavBar.js` for the major routing-related changes. After that it's all small changes and re-organizing. 

It'd probably be easier to review and give feedback on the folder changes by just checking it out locally and poking around in the file tree. 

Testing should just be clicking around making sure you can and can't access pages that you should or shouldn't be able to :)

### Related PRs

Branches off of #65 

CC: @dfangshuo


src/components/AuthenticatedRoute.js
```diff
@@ -0,0 +1,50 @@
...
+      authed = userCredentials.includes(credential);
+    } else if (noauth) {
+      authed = !isSignedIn(userCredentials);
+    } else {
+      authed = isSignedIn(userCredentials);
+    }
+
+    // TODO: If you visit a non-authenticated route, it should show you an error or something.
+    // Right now it just redirects to the homepage
+    // Alternatively, we can render a generic 404 component if they reach somewhere they shouldn't,
```
dfangshuo: hmm personally i feel like redirecting to the homepage is a fine course of action

src/components/AuthenticatedRoute.js
```diff
@@ -0,0 +1,50 @@
...
+    let authed;
+    // If credential prop exists, ensure they are authorized
+    // If noauth prop exists, ensure they are NOT signed in
+    // else, just ensure they are signed in
+    if (credential) {
+      authed = userCredentials.includes(credential);
+    } else if (noauth) {
+      authed = !isSignedIn(userCredentials);
+    } else {
+      authed = isSignedIn(userCredentials);
```
dfangshuo: this authentication code block would be a good candidate to put in a helper function ( `credential => authed` )

src/components/NavBar.js
```diff
@@ -10,52 +10,11 @@ import {
...
-
 class NavBar extends React.PureComponent {
   render() {
-    const { person, credentials, path } = this.props;
-    // TODO: Holy crap so apparently the "ID" column on airtable is their NAME not the record ID.
-    // we NEED to standardize airtable column names so things aren't hella confusing
+    const { person, credentials } = this.props;
     const displayName = person ? person.name : '';
 
-    if (path === ONBOARDING_ROUTE) {
```
dfangshuo: im not sure the Navbar behaves as expected anymore after this change. Before, we had 3 states for the Navbar: `onboarding`, `!isSignedIn`, and regular, now it seems like there's only 1 state? more importantly, the navbar at the login screen doesn't have anything in the navbar

image

... which is fine, but the thing is when i try to sign up, there's no way for me to go back to the login screen now, too

image

src/App.js
```diff
@@ -7,36 +7,84 @@ import { ConnectedRouter } from 'connected-react-router';
...
+            <AuthenticatedRoute path="/community" component={Community} />
+            <AuthenticatedRoute path="/profile" component={UserProfile} />
+
+            <AuthenticatedRoute
+              noauth // Signed out users only
+              path="/onboarding"
+              component={Onboarding}
+            />
+            <AuthenticatedRoute
+              credential="A" // Admins only
```
dfangshuo: hmm if we're actually moving forward with this string parsing approach, perhaps we can put these letters into `constants.js` for more robustness? Or if this is not the approach we want to ultimately move forward with we should put a todo or something

src/screens/subscriber/components/AllBillsTable.js
```diff
@@ -62,7 +62,7 @@ const createFullBillTransaction = transaction => {
   };
 };
 
-export default class SubscriberOwnerDashboardAllBillsTable extends React.Component {
+export default class SubscriberAllBillsTable extends React.Component {
```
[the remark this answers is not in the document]
aivantg: YES definitely. this is actually just because i kept changing the names and forgot to change some of the file names

src/screens/subscriber/Billing.js
```diff
@@ -47,13 +41,13 @@ class SubscriberOwnerDashboard extends React.Component {
     }
   }
 
-  seeSubscriberOwnerDashboardAllBillsView() {
+  seeAllBillsView() {
```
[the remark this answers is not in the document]
aivantg: sure

src/screens/auth/Login.js
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
-import { connect } from 'react-redux';
 import { loginUser } from '../../lib/authUtils';
 import '../../styles/Login.css';
 import '../../styles/main.css';
 
-const HOME_ROUTE = '/dashboard';
+const HOME_ROUTE = '/';
```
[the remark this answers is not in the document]
aivantg: yeah that seems like a good idea!

src/App.js
```diff
@@ -7,36 +7,84 @@ import { ConnectedRouter } from 'connected-react-router';
...
+} from './lib/credentials';
+import AuthenticatedRoute from './components/AuthenticatedRoute';
+import Investment from './screens/general/Investment';
 
 class App extends React.Component {
   componentDidMount() {
     const { userLogin } = this.props;
+
+    // If userLogin info is in Redux, fetch latest version
     if (userLogin) {
```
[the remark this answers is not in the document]
aivantg: No, it's the full record. I am thinking about potentially changing how it works but no you're supposed to pass in the full record, which is what is stored in redux

src/components/AuthenticatedRoute.js
aivantg: mm, maybe, but it requires data that comes from the props of this component to realistically compute what's necessary. Maybe I could put it in a helper function within the component though!

src/App.js
aivantg: yeah..i was thinking about this. will come up with a solution for this.

src/components/NavBar.js
aivantg: This was actually the intended design since I thought it was weird to have the sign in button be present throughout onboarding, and given the way that onboarding works that would be extra difficult. I figured they could just go back to the homescreen if they wanted to go back to sign ni (or maybe a button on the actual onboarding page would be more fitting.

src/App.js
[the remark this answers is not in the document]
dfangshuo: ok. Is there any way this might be documented to make it more clear, cause i can't seem to find this stated anyway

src/lib/credentials.js
```diff
@@ -3,6 +3,8 @@ import constants from '../constants';
 
 const { SUBSCRIBER_OWNER, GENERAL_OWNER } = constants;
 
+const Credentials = { ADMIN: 'A', GENERAL: 'G', SUBSCRIBER: 'S' };
```
dfangshuo: hmm is there a reason why we capitalize this :o feels a little contrary to convention

aivantg: Hm, honestly it was sort of just an instinct tbh. I'm used to accessing constants from an object using capitalized names (which i realize I borrowed from Java because constants are stored in it's own package oftentimes) and then the actual name in all caps. 

I'm biased but up to you

aivantg:
I'll see if I can clear up redux stuff later

PR #67, opened 2020-01-28
Cleaned up Admin Dashboard + Utils

aivantg:

## What's new in this PR?

I cleaned up the logic in admin utils to make a lot more sense. In short, admin dashboard gets the owner ID's from the project group record (stored in redux) and pulls the owner records once when the components mount and any time that project group record is changed. It then passes the full owner record to each `AdminDashboardCard` instance which makes calls to `removeOwner` (which has been pushed to `adminUtils.js`

This PR also fixes a small bug in how admin credentialing is calculated

### How to Review

Normal Review. Removing admins should work. 

### Related PRs

Merges into #66 

I did not add tests for this.....mostly cuz i'm lazy but also deprioritizing it. let me know if you wanna come back and add them. 

### Screenshots

CC: @dfangshuo


src/lib/adminUtils.js
```diff
@@ -1,35 +1,20 @@
...
-
-const updateProjectGroupOwners = async (groupId, newOwners) => {
-  return updateProjectGroup(groupId, {
+  await updateProjectGroup(projectGroup.id, {
     owner: newOwners
   });
-};
 
-export { getAdminTable, getOwnersFromProjectGroup, updateProjectGroupOwners };
+  // Download latest data, including new project group info
```
dfangshuo: hmm, this comment is a little ambiguous, maybe something like: 'refresh local copy of the data after the owner has been removed' or something along those lines might be more helpful/descriptive?

src/lib/credentials.js
```diff
@@ -1,20 +1,18 @@
-import { getAdminTable } from './adminUtils';
 import constants from '../constants';
 
 const { SUBSCRIBER_OWNER, GENERAL_OWNER } = constants;
 
 const Credentials = { ADMIN: 'A', GENERAL: 'G', SUBSCRIBER: 'S' };
```
dfangshuo: Think I raised this in the #66, but in case I realized it might be more relevant to address it here: is there a reason why `Credentials` is capitalized? Feels a little like it's going against our convention

src/lib/credentials.js
```diff
@@ -1,20 +1,18 @@
...
   let credentials = '';
 
   // TODO: Will this ever be the case? Is this check necessary
   if (owner == null) {
     return credentials;
   }
 
-  const isAdminPayload = await getAdminTable(owner);
-  if (isAdminPayload !== -1) {
+  if (owner.adminOf && owner.adminOf.length >= 0) {
```
dfangshuo: hmm i guess this wouldn't handlle the (edge) case of an owner being the admin of multiple project groups? or can that never happen

src/screens/admin/AdminDashboard.js
```diff
@@ -2,83 +2,44 @@ import React from 'react';
...
-    const { adminGroupId, adminId, owners } = this.state;
-    let ownerIds = owners.map(owner => owner.ownerId);
-
-    // FILTER
-    ownerIds = ownerIds.filter(ownerId => ownerId !== idToRemove);
-    const newOwners = owners.filter(owner => owner.ownerId !== idToRemove);
-    ownerIds.push(adminId);
-
-    await updateProjectGroupOwners(adminGroupId, ownerIds);
+  async fetchOwnerRecords() {
```
dfangshuo: this should be in `adminUtils`. Putting this function here would break the abstraction barrier that we've setup. That function could be one that takes in `projectGroup` and then returns `ownerRecords`

src/screens/admin/components/AdminDashboardCard.js
```diff
@@ -1,33 +1,45 @@
...
-  const ownersTags = ownerType.map(type => {
-    return (
+class AdminDashboardCard extends React.PureComponent {
+  render() {
+    const { owner } = this.props;
+    const { ownerType: ownerTypes } = owner;
+    const credentials = getCredentials(owner);
+    const admin = isAdmin(credentials);
+    if (admin) {
+      ownerTypes.push('Admin');
```
dfangshuo: maybe declare this as a constant either at the top of the file or in `constants.js`?

src/screens/admin/components/AdminDashboardCard.js
```diff
@@ -1,33 +1,45 @@
...
 
-function AdminDashboardCard({ name, callback, ownerId, ownerType }) {
-  const ownersTags = ownerType.map(type => {
-    return (
+class AdminDashboardCard extends React.PureComponent {
+  render() {
+    const { owner } = this.props;
+    const { ownerType: ownerTypes } = owner;
+    const credentials = getCredentials(owner);
+    const admin = isAdmin(credentials);
```
dfangshuo: hmm maybe something like `hasAdminPermissions` would be more descriptive?

src/screens/admin/components/AdminDashboardCard.js
```diff
@@ -1,33 +1,45 @@
...
+    return (
+      <div className="admin-card">
+        <div className="card-profile-image" />
+        <div className="card-name">
+          {/* We can use their ID here because their id is...technically their name */}
+          <h3>{owner.id}</h3>
+          <div className="card-tags">{ownerTags}</div>
+        </div>
+        <div>
+          {!admin && (
```
dfangshuo: At first I was a little confused by this, I think it would be helpful to put here that this ensures that only non-admins can be removed from the group

src/lib/credentials.js
aivantg: Again, it's a convention that I'm used to from java (accessing constants from an object is usually capitalized)...i don't know if there's convention for it in javascript...will look into

src/lib/credentials.js
aivantg: Yes. I thought I noted this somewhere but the current code makes that assumption which I think is okay for now. Worth noting to clarify with grayson though now that I think about it.

src/screens/admin/AdminDashboard.js
aivantg: Which abstraction barrier would it break? I'm down to move it if you want to establish that no screens have airtable logic, but wanted to make sure (for the time being, will move)

src/screens/admin/components/AdminDashboardCard.js
aivantg: good idea, was trying to think of a better name

src/screens/admin/components/AdminDashboardCard.js
aivantg: I don't know if that's worth it since this is only related to the way that Admin is denoted in this unique part of the app. Could put it at the top of our app, but that would also suggest we should put _all_ copy in our app in separate files...which is probably ideal but tedious? 

Also, this doesn't follow designs yet so this might actually be revisited and redone. idk, thoughts?

src/screens/admin/components/AdminDashboardCard.js
aivantg: 👍

src/lib/credentials.js
dfangshuo: maybe we can leave a comment here (about this assumption) just so this is documented? I'll also add it to a list of questions to ask Grayson

src/screens/admin/AdminDashboard.js
dfangshuo: oh yes! to clarify: functions from `request.js` ideally shouldn't be used in any of the screens, as any use of functions from `request.js` should be wrapped in a function in a `___Utils.js` file. Screen should call the admin function instead. [Notion reference here](https://www.notion.so/PP-Power-Airtable-Client-Side-6a1b6734af294ef88609a6d6d256ca3d)

src/screens/admin/components/AdminDashboardCard.js
dfangshuo: ah i see, this is the 'Admin' text that is displayed as copy - makes sense! I think i misunderstood what this is supposed to do

src/screens/admin/AdminDashboard.js
aivantg: I have since updated this!

src/lib/credentials.js
aivantg: I believe I have added a comment, don't know why it's not showing up here

PR #78, opened 2020-02-28
Investments Webpage

cindyzhang12:

## What's new in this PR?

Created the title and boxes for the Investments page

### How to Review

Review Investments.css and Investment.js
(just realized the css file has Investments and js file is Investment, we should change the name so it's consistent!)

### Related PRs

N/A

### Next steps

Make the code prettier and maybe more succinct, I had a hard time figuring out the centering and I'm sure there probably a better way to do it than I did

### Screenshots

CC: @dfangshuo


niwong:
Hey! Nice job laying out all the boxes, it looks good when I ran it. We can go over it on Wednesday, and I will leave some comments on your CSS.

src/screens/general/Investment.js
```diff
@@ -1,8 +1,35 @@
 import React from 'react';
+import '../../styles/Investments.css';
 
 class Investment extends React.Component {
+  constructor(props) {
```
aivantg: In this component, since we’re not actually using state anywhere (yet), we actually don’t need to bother initializing it at all! So you could get rid of this constructor entirely for this component

aivantg: In this component, since we’re not actually using state anywhere (yet), we actually don’t need to bother initializing it at all! So you could get rid of this constructor entirely for this component

src/styles/Investments.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.everything h1 {
+  margin-top: 40px;
+  font-style: normal;
+  font-weight: 900;
+  font-size: 36px;
+  line-height: 49px;
+  color: var(--pp-blue);
+}
```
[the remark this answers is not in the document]
aivantg: So cool ^ 😮 

Another comment/question I’d add, we might want to use a more descriptive name for the “everything” class so that it’s easy to come back to our CSS and remember what this applies to!

src/styles/Investments.css
```diff
@@ -0,0 +1,77 @@
...
+
+.everything {
+  flex-direction: column;
+}
+
+.leftandright-content {
```
aivantg: Another place where we might want to think of a more expressive name!

src/screens/general/Investment.js
aivantg: If ESLint yells at you (Doesn’t let you commit) when you remove the constructor, I think you can get it to shut up by replace `React.Component` to `React.PureComponent`. Can explain more on Sunday!

src/styles/Investments.css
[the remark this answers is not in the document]
cindyzhang12: Oh that's cool!! I didn't know we could view her Figma designs like that, I was pretty much eyeballing everything from Fang's screenshot but I will def use that in the future!! :)

src/styles/Investments.css
```diff
@@ -0,0 +1,73 @@
+.dashboard {
```
aivantg: Super small nit actually, could you remove `.dashboard`. Since this is already defined in `App.css`, there's no need to redefine it :)

PR #79, opened 2020-02-29
New Favicon

cindyzhang12:

## What's new in this PR?

Added new favicon to the webpage!

### How to Review
- Changed code in index.html and manifest.json
- Uploaded .png and .ico versions of Iris' logo

### Related PRs
N/A

### Next steps
Wasn't super sure about the sizes of the images - I just downloaded the png from Notion and used that for every one, wasn't sure if I had to resize to the size specified in the code.

### Screenshots

CC: @dfangshuo


src/styles/Investments.css
```diff
@@ -0,0 +1,77 @@
+.cont {
```
dfangshuo: Is there a reason this file is also in this PR? If it's not related it should not be in this PR

src/App.js
```diff
@@ -64,6 +64,11 @@ class App extends React.Component {
     return (
       <ConnectedRouter history={history}>
         <div className="app-container">
+          <link
```
dfangshuo: hmm what what is this link for?

public/manifest.json
```diff
@@ -3,17 +3,17 @@
   "name": "Create React App Sample",
   "icons": [
     {
-      "src": "favicon.ico",
+      "src": "ppsc-favicon-01.ico",
```
dfangshuo: is there a reason this is ppsc-favicon-01 instead of just ppsc-favicon? If not I would recommend just naming it ppsc-favicon

cindyzhang12: Iris' file on her sprint task was ppsc-favicon-01 so I kept it that way because I thought it was named that for a reason! If not, I can change it

dfangshuo: I think that's a good enough reason - let's keep it as this!

src/App.js
cindyzhang12: updated, this part was deleted

PR #89, opened 2020-03-05
Prefill Onboarding with invite link

dfangshuo:

## What's new in this PR?

[//]: # "Describe what's new in this PR"
Modify onboarding to accept the ability to prefill forms
- Pledge Invite is deleted upon successful onboarding (when the invitee gets to the last page of onboarding)

### How to Review

[//]: # 'The order in which to review files and what to expect when testing locally'

### Related PRs

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc"

### Next steps

[//]: # "What doesn't work yet, what's NOT in this PR/has to be done "

### Screenshots

CC: @dfangshuo

### DIFF
--- a/src/lib/onboardingUtils.js
+++ b/src/lib/onboardingUtils.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-await-in-loop */
 import States from '../assets/states.json';
 import {
+  deletePledgeInvite,
   getOwnersByEmail,
   getAllProjectGroups,
   getPledgeInviteById,
@@ -137,9 +138,20 @@ const safeGetPledgeInviteById = async pledgeInviteId => {
   }
 };
 
+const safeDeletePledgeInvite = async pledgeInviteId => {
+  try {
+    const pledgeInvite = await deletePledgeInvite(pledgeInviteId);
+    return pledgeInvite;
+  } catch (e) {
+    console.log(e); // TODO: more granular error handling
+    return null;
+  }
+};
+
 export {
   validateField,
   getAvailableProjectGroups,
+  safeDeletePledgeInvite,
   safeGetPledgeInviteById,
   updateOwnerFields
 };
--- a/src/lib/onboardingUtils.js
+++ b/src/lib/onboardingUtils.js
@@ -3,6 +3,7 @@ import States from '../assets/states.json';
 import {
   getOwnersByEmail,
   getAllProjectGroups,
+  getPledgeInviteById,
   updateOwner,
   createOwner
 } from './airtable/request';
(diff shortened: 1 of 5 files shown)


src/screens/onboarding/steps/ContactInfoStep.js
```diff
@@ -4,7 +4,10 @@ import '../../../styles/main.css';
 
 class ContactInfoStep extends React.PureComponent {
   render() {
-    const { owner, errors, onSubmit, handleChange } = this.props;
+    const { owner, errors, onSubmit, handleChange, pledgeInvite } = this.props;
+    if (pledgeInvite) {
+      owner.phoneNumber = pledgeInvite.phoneNumber;
```
aivantg: This way of passing in pledge invite data is alright, but one thing that it misses is that, if the user refreshes the page and comes back to onboarding later, the pledge invite will be lost. Aka with this flow, they have to fill out _all_ of onboarding at once or they can't use the invite. We should do our best to mesh the invite flow with our regular flow. 

When the user fills out the Basic Info step, we create their entire owner record, filling in their name, email, and password. _Here_ is when we should add the extra records the pledge invite provides us with! Lmk if this doesn't make sense. I'll comment where this happens

src/screens/onboarding/Onboarding.js
```diff
@@ -49,7 +71,7 @@ class Onboarding extends React.Component {
   };
 
   nextStep = async event => {
-    const { owner } = this.state;
+    const { owner, pledgeInvite } = this.state;
     if (event) {
```
aivantg: I think the best way to do this would be in _this_ function, if step == 1 add the pledge invite fields to the `newOwner` object. I would skip validation (because they would not be able to correct anything if like something later down the line was invalid) and just add it manually. Let me know if this makes sense!

src/screens/onboarding/Onboarding.js
```diff
@@ -85,6 +107,14 @@ class Onboarding extends React.Component {
         newOwner.projectGroupId = [newOwner.projectGroupId];
       }
 
+      // delete the pledge invite if it exists on success
+      if (owner.onboardingStep === 5 && pledgeInvite) {
```
aivantg: again, I'd just add the fields to `newOwner` here (after the basic info step, can't remember if it's step 0 or 1) and delete the pledge invite after. Pledge invite should be tossed after they create an account. 

Also, something that we should maybe consider is skipping the project group screen, as if the invite locks them into a project group. Otherwise, if they get invited to a private project group, they could select a public one and never get access to the private one :/ this might require a flag of some kind on the owner object to signify that the project group they're in is locked in.

src/lib/onboardingUtils.js
```diff
@@ -126,4 +128,30 @@ const updateOwnerFields = async (owner, fields) => {
   }
 };
 
-export { validateField, getAvailableProjectGroups, updateOwnerFields };
+const safeGetPledgeInviteById = async pledgeInviteId => {
```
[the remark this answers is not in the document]
dfangshuo: Hmm what would the 'token' refer to in this case?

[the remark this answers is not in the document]
aivantg: token being the invite "code". Which in this case happens to be the pledge invite ID

lol i'm not picky in terms of the name, was just a thought instead of "safe". i'm okay with whatever!

PR #85, opened 2020-03-03
Ashley/modal

ashleynguyen27:
[//]: # "These comments meant for your reference, they are invisible and don't need to be deleted"

## What's new in this PR?
Created a simple Modal component
[//]: # "Describe what's new in this PR"
A new modal component was made.
### How to Review
checkout localhost:3000/workshop/modal to see the example
[//]: # 'The order in which to review files and what to expect when testing locally'

### Related PRs

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc"

### Next steps

[//]: # "What doesn't work yet, what's NOT in this PR/has to be done "

### Tests Performed, Edge Cases

[//]: # 'If you made changes/additions to `airtable.js` make sure you update/add'
[//]: # '[at least 1] corresponding unit tests in `airtable.spec.js`'

### Screenshots
Screen Shot 2020-03-03 at 12 43 34 AM (2)

[//]: # "Add screenshots!!! If you'd like"

CC: @dfangshuo

PR #90, opened 2020-03-05
Tweaked modal using react-modal

dfangshuo:

## What's new in this PR? 
[//]: # "Describe what's new in this PR"
Modified the general purpose modal that @ashleynguyen27 implemented in Sprint #2 with the `react-modal` library.

Also renamed `Modal.js` to `PpModal.js` to distinguish it from the `Modal` from `react-modal`
### How to Review 
[//]: # 'The order in which to review files and what to expect when testing locally'
First check out `PpModal`, which is the re-implementation of the Modal using `react-modal`.

Then, look at `ModalDemoScreen` (previously `ModalDemo`) to see how it's used, paying particular attention to the `showModal` state management 

### Related PRs

[//]: # "Optional - related PRs you're waiting on/ PRs that will conflict, etc"
Merge this PR into https://github.com/calblueprint/peoplepower-web/pull/85 before merging https://github.com/calblueprint/peoplepower-web/pull/85

### Next steps

[//]: # "What doesn't work yet, what's NOT in this PR/has to be done "

### Screenshots

image

CC: @dfangshuo


src/components/PpModal.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import Modal from 'react-modal';
+import '../styles/Modal.css';
+
+const ROOT_ELEMENT = '#root';
+Modal.setAppElement(ROOT_ELEMENT);
+
+export default class PpModal extends React.PureComponent {
```
aivantg: I think convention would say it should be named `PPModal` instead of `PpModal` since PP stands for PeoplePower which are two separate words. Also just looks nicer

PR #92, opened 2020-03-06
General Owner Dashboard Updates

cindyzhang12:

## What's new in this PR?
[//]: # "#####################################"
[//]: # "####### YOUR ANSWER BELOW ###########"
- Inserted arrow next to dashboard header
- Created My Investment section along with arrow

[//]: # "############## END ##################"

### How to Review
[//]: # "#####################################"
[//]: # "####### YOUR ANSWER BELOW ###########"

- look into generaldashboard.js and generaldashboard.css

[//]: # "############## END ##################"

### Related PRs
[//]: # "#####################################"
[//]: # "####### YOUR ANSWER BELOW ###########"

[//]: # "############## END ##################"

### Next steps
PR/has to be done "
[//]: # "#####################################"
[//]: # "####### YOUR ANSWER BELOW ###########"
- Clarify with Iris about spacing between solar projects and investment headers, couldn't get the margins to show up on Figma

[//]: # "############## END ##################"

[//]: # "#####################################"
[//]: # "####### YOUR ANSWER BELOW ###########"

[//]: # "############## END ##################"

### Screenshots
[//]: # "#####################################"
[//]: # "#### YOUR SCREENSHOTS BELOW ########"

[//]: # "############## END ##################"
CC: @dfangshuo


src/screens/general/GeneralDashboard.js
```diff
@@ -4,6 +4,7 @@ import { logOut } from '../../lib/authUtils';
 import AnnouncementList from '../shared/components/AnnouncementList';
 import LoadingComponent from '../../components/LoadingComponent';
 import '../../styles/GeneralOwnerDashboard.css';
+import RightArrow from '../../assets/see more.png';
```
aivantg: I'd highly recommend _against_ naming files with spaces in them. use underscores, dashes, or just nothing to combine the words instead. Spaces in names can cause a lot of trouble some times when dealing with files on the terminal

aivantg: You also might want to name the file something more descriptive like maybe "see_more_arrow.png" so that someone knows what it is by looking at the filename

src/screens/general/GeneralDashboard.js
```diff
@@ -71,7 +72,16 @@ class GeneralOwnerDashboard extends React.Component {
     return (
       <div className="dashboard">
         <div className="cont dash-announcements-cont">
-          <h3>Project News</h3>
+          <div className="headerbutton">
+            <div className="headeronly">
+              <h3>Project News</h3>
+            </div>
+            <div className="rightbutton">
+              <a href="/projectnews">
```
aivantg: Could you switch the `<a/>` tags to be `<Link />` tags instead (like in `Navbar.js`)?

The difference is subtle, but using the `Link` component keeps the information in redux (our global store) updated. You'll notice that right now, if you follow one of the arrows, it'll take you to the right page, but the highlighted page on the navbar will be inaccurate. 

This should also make the transition between pages faster! Happy to explain why in person if you're curious 😄

src/screens/general/GeneralDashboard.js
cindyzhang12: makes sense! i just used the name that was on the file when i downloaded it, but ill fix it! :)

PR #99, opened 2020-03-11
Ashley/subscriber dashboard

ashleynguyen27:

## What's new in this PR?
Added a subscriber dashboard that can be used for subscribers and subscriber owners. This dashboard includes a billing, project news, and charts (not finished in this pr) sections. 

### How to Review
Check the SubscriberDashboard.js file.
This file was too big so it was split into sections. The name of the components are: 
- DashboardBillingSection
- DashboardChartsSection
- DashboardProjectNewsSection. 
Billing.js and Navbar.js were also modified. 

Demo: Log into sam@berkeley.edu to see an example subscriber dashboard. 

- SubscriberDashboard.js
- DashboardBillingSection
- DashboardChartsSection
- DashboardProjectNewsSection 
- Billing.js
- Navbar.js

Demo: Log into sam@berkeley.edu to see an example subscriber dashboard. The Subscriber board should look like the prototype Iris made 

### Related PRs

### Next steps

PR/has to be done "

### Screenshots

[
Screen Shot 2020-03-12 at 9 28 46 PM
](url)

CC: @dfangshuo


src/screens/subscriber/Billing.js
```diff
@@ -55,27 +55,26 @@ class Billing extends React.Component {
 
   render() {
     const { mode, transactions, isReady, pendingBills } = this.state;
-    const { isLoadingUserData } = this.props;
+    const { isLoadingUserData, location } = this.props;
+    const { mode2 } = location;
```
aivantg: I see what you did here and it's pretty sweet, but I think a slightly cleaner way to do it, is instead of having two `mode` parameters and comparing based on both of them, you can keep the mode in a query string! 

So when you link to the billing page, instead of linking to `/billing`, link to `/billing?view=all`

and then in the constructor, you can pull out the value of `view` and set the initial state for `mode` based on that. I just recently added a library to help do that called `qs`. 

```
// Top of file
import qs from 'qs';

// In Constructor
const {location} = this.props
const view = qs.parse(location.search, {
  ignoreQueryPrefix: true
}).view;

if (view === 'all') { 
  this.state.mode = 1 // It's okay to edit state directly because this code would be in the constructor
}
```

This is actually important because otherwise, you would never be able to navigate away from the AllBillsview

src/components/NavBar.js
```diff
@@ -47,7 +47,14 @@ class NavBar extends React.PureComponent {
                     pathname === '/billing' ? 'navItemSelected' : 'navItem'
                   }
                 >
-                  <Link to="/billing">Billing</Link>
+                  <Link
+                    to={{
+                      pathname: '/billing',
+                      mode2: 0
```
aivantg: following with comment below, remove mode2 and this edit, instead use query parameters

src/screens/subscriber/SubscriberDashboard.js
```diff
@@ -1,9 +1,159 @@
...
+    this.setState({
+      mode: 0
+    });
+  }
+
+  switchChartView() {
+    const { activeTab, hasShares } = this.state;
+    if (activeTab === 'My Solar Project' && hasShares) {
+      this.setState({
+        activeTab: 'Community Solar Projects'
```
aivantg: I def am not perfect about this either, but i think these should be constants. since the names are so long, we're one typo away from a very hard to find bug LOL

src/screens/subscriber/components/DashboardBillingSection.js
```diff
@@ -0,0 +1,138 @@
...
+  render() {
+    const { totalBalance, data } = this.props;
+    return (
+      <div>
+        <div className="subscriber-section-header">
+          <div className="subscriber-header">Billing Summary</div>
+          <Link
+            to={{
+              pathname: '/billing',
+              mode2: 0
```
aivantg: I think this one should be just `/billing` (with query param update)

src/screens/subscriber/components/DashboardBillingSection.js
```diff
@@ -0,0 +1,138 @@
...
+                    type="button"
+                    className={`subscriber-billing-make-payment-button ${
+                      totalBalance === 0 ? 'disabled' : ''
+                    }`}
+                    disabled={totalBalance === 0}
+                  >
+                    <Link
+                      to={{
+                        pathname: '/billing',
+                        mode2: 1
```
aivantg: here we should pass in the query param `/billing?view=all`

src/screens/subscriber/components/DashboardProjectNewsSection.js
```diff
@@ -0,0 +1,47 @@
...
+          <div className="subscriber-header">Project News</div>
+          <Link to="/projectnews">
+            <img src={RightArrow} alt="right arrow" />
+          </Link>
+        </div>
+        <div className="subscriber-side-section-body">
+          {isLoadingAnnouncements ? (
+            <LoadingComponent />
+          ) : (
+            announcements.map(announcement => {
```
aivantg: I think ideally we make some sort of a reusable component across the many times that community cards are displayed...but this can be something we talk about in cleanup rather than something that falls under _this_ PR. Note for @dfangshuo

src/styles/SubscriberDashboard.css
```diff
@@ -0,0 +1,184 @@
+.subscriber-page {
+  margin: 1rem 7rem;
```
niwong: Ohhh man, should I be using rems? Do we have a standard for that?

src/screens/subscriber/Billing.js
niwong: Oh that's a super cool idea.

src/styles/SubscriberDashboard.css
ashleynguyen27: tbh idk what is standard but i started using rem in the beginnning. i regret no one else uses it

dfangshuo:
@ashleynguyen27  Follow the chronology of the Project News page (aka latest announcements should be on top). It should be a simple reverse list function (see how project news does it)

src/screens/subscriber/components/DashboardProjectNewsSection.js
ashleynguyen27: ya, I was actually going to delete in the meantime cause I hated seeing it for the millisecond it was on the screen.

src/screens/subscriber/components/DashboardBillingSection.js
ashleynguyen27: ohhhhh tyty

src/screens/subscriber/components/DashboardChartsSection.js
```diff
@@ -0,0 +1,72 @@
...
+      activeTab: 0
+    };
+    this.switchChartView = this.switchChartView.bind(this);
+  }
+
+  // need to make sure the value doesnt switch if its the same button
+  switchChartView = e => {
+    const { activeTab } = this.state;
+    const { hasShares } = this.props;
+    if (e.target.value != 0 && activeTab === 0 && hasShares) {
```
dfangshuo: if you want to check that `e.target.value` is not null, just do 
```suggestion
    if (e.target.value && activeTab === 0 && hasShares) {
```

!= should be avoided as much as possible

src/screens/subscriber/components/DashboardChartsSection.js
```diff
@@ -0,0 +1,72 @@
...
+      this.setState({
+        activeTab: 1
+      });
+    } else if (e.target.value != 1) {
```
dfangshuo: similarly: 
```suggestion
    } else if (e.target.value) {
```

dfangshuo:
Use 
```
switchChartView = e => {
    const { hasShares } = this.props;
    if (!hasShares) {
      return;
    }

    // const { activeTab } = this.state;
    const toSelect = parseInt(e.target.value);

    this.setState(prevState => {
      const { activeTab: prevActiveTab } = prevState;
      
      if (prevActiveTab === toSelect) {
        return {};
      }

      return { activeTab : toSelect };
    });
  };

```

for the `!=` part

PR #104, opened 2020-03-14
Investments Webpage Box Content and Modal

cindyzhang12:

## What's new in this PR?

### How to Review

- Read through Investments.js and Investments.css 
- Implemented the modal for changing dividend preferences, which can be found in the files mentioned above
### Related PRs

- Need to get this approved so I can start working on the buying shares pages!
### Next steps
- Currently the buy shares button does not work yet and will just redirect to the same page since it hasn't been implemented yet!
PR/has to be done "

### Screenshots

CC: @dfangshuo


package.json
```diff
@@ -14,6 +14,7 @@
     "google-maps-react": "^2.0.2",
     "history": "^4.10.1",
     "react": "^16.10.1",
+    "react-circular-progressbar": "^2.0.3",
```
dfangshuo: actually did not know this existed lmao... nice

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
   height: 250px;
   display: flex;
 }
+
+.buy-shares-button {
+  font-family: 'Nunito Sans', sans-serif;
+  justify-content: center;
+}
```
dfangshuo: this is too general a name. Maybe call it something like investments-buy-shares-button to prevent css collision

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
...
+
+.dividend {
+  justify-content: center;
+  align-content: center;
+}
```
dfangshuo: ditto to above: change it to investsments-dividend-box or something

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
...
+
+.buttons {
+  display: flex;
+  flex-direction: column;
+  text-align: center;
+  margin-right: -10px;
+}
```
dfangshuo: ditto. this one is the most general (need to change)

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
...
+
+.investment-box-2 h4 {
+  text-align: left;
+  color: var(--pp-blue);
+}
+
+.preferences {
+  display: flex;
+  flex-direction: column;
+}
```
dfangshuo: too general

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
...
+  text-decoration: none;
+  display: inline-block;
+  font-size: 16px;
+  width: 150px;
+}
+
+.investment-box-2 h4 {
+  text-align: left;
+  color: var(--pp-blue);
+}
```
dfangshuo: what does 2 mean in this case? maybe more descriptive naming

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
...
+.buy-shares-button a {
+  background-color: var(--pp-pink);
+  border-radius: 2.5px;
+  color: white;
+  text-align: center;
+  text-decoration: none;
+  display: inline-block;
+  font-size: 16px;
+  width: 150px;
+}
```
dfangshuo: investments-buy-shares-button or something

src/styles/Investments.css
```diff
@@ -64,3 +70,99 @@
...
+  display: flex;
+  flex-direction: column;
+}
+
+.preferences h4 {
+  color: var(--pp-black);
+  margin-block-start: 0;
+  margin-block-end: 0;
+  margin-bottom: 8px;
+}
```
dfangshuo: ditto for everything below. Make naming more specific

src/screens/general/Investment.js
```diff
@@ -1,17 +1,83 @@
...
+                  <CircularProgressbar
+                    viewBox="0 0 0 0"
+                    value={percentage}
+                    text={owner.numberOfShares}
+                    styles={buildStyles({
+                      pathColor: '#cd6795',
+                      textSize: '30px',
+                      textColor: '#cd6795',
+                      trailColor: '#F4F1F24'
+                    })}
```
dfangshuo:
```suggestion
                    styles={buildStyles(circularProgressBarStyles)}
```

Then, define (at the top of this file):

```
const circularProgressBarStyles = {
                      pathColor: '#cd6795',
                      textSize: '30px',
                      textColor: '#cd6795',
                      trailColor: '#F4F1F24'
                    }
```

src/screens/general/Investment.js
```diff
@@ -1,17 +1,83 @@
...
 class Investment extends React.PureComponent {
   render() {
+    const { owner } = this.props;
+    const percentage = owner.numberOfShares * 10;
     return (
       <div className="dashboard">
         <div className="mainheader">
           <h1>My Investment</h1>
           <div className="columnformat">
             <div className="left-content">
```
dfangshuo: what is this class (left-content)? I don't see it defined anywhere

src/styles/Investments.css
dfangshuo: e.g. something like investment-box-top or -bottom would be more descriptive

src/screens/general/Investment.js
```diff
@@ -1,17 +1,196 @@
...
+
+  onSavePreferencesPressed = async () => {
+    const { newIsReceivingDividends } = this.state;
+    this.submitPreference({
+      isReceivingDividends: newIsReceivingDividends
+    });
+  };
+
+  onClickSavePreferences = () => {
+    this.onSavePreferencesPressed();
```
dfangshuo:
```suggestion
    await this.onSavePreferencesPressed();
```

src/screens/general/Investment.js
```diff
@@ -1,17 +1,196 @@
...
+
+  submitPreference = async newIsReceivingDividends => {
+    const { owner } = this.props;
+    await updateOwner(owner.id, newIsReceivingDividends);
+    await refreshUserData(owner.id, owner.isReceivingDividends);
+  };
+
+  onSavePreferencesPressed = async () => {
+    const { newIsReceivingDividends } = this.state;
+    this.submitPreference({
```
dfangshuo:
```suggestion
    await this.submitPreference({
```

src/screens/general/Investment.js
```diff
@@ -1,17 +1,196 @@
...
+
   render() {
+    const { owner } = this.props;
+    const { showModal, newIsReceivingDividends } = this.state;
+    const percentage = owner.numberOfShares * 10;
+    const circularProgressBarStyles = {
+      pathColor: '#cd6795',
+      textSize: '30px',
+      textColor: '#cd6795',
+      trailColor: '#F4F1F24'
```
dfangshuo: define these colors at the top of the file, and call them here. E.g. define

```
const SOME_COLOR = '$cd6795'

...

pathColor: SOME_COLOR

File: src/styles/Investments.css [hunk 14]
```diff
@@ -64,3 +70,150 @@
...
+  transform: scale(0.9);
+}
+
+.dividends-modal {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  border: 0px;
+  background: rgb(255, 255, 255);
```
dfangshuo:
```suggestion
  background: white;
```

src/styles/Investments.css
```diff
@@ -64,3 +70,150 @@
...
+  background-color: var(--pp-black);
+  border-radius: 40px;
+  color: white;
+  text-align: center;
+  text-decoration: none;
+  font-size: 16px;
+  width: 100px;
+  height: 25px;
+  align-content: flex-end;
+  font-family: 'Nunito Sans', sans-serif;
```
dfangshuo: same comment as above for fonts

src/styles/Investments.css
```diff
@@ -64,3 +70,150 @@
   height: 250px;
   display: flex;
 }
+
+.investments-buy-shares-button {
+  font-family: 'Nunito Sans', sans-serif;
```
dfangshuo: is there a reason you are using a custom font here? You shouldn't need to define any type of fonts in these css files

src/screens/general/Investment.js
```diff
@@ -59,11 +59,15 @@ class Investment extends React.PureComponent {
     const { owner } = this.props;
     const { showModal, newIsReceivingDividends } = this.state;
     const percentage = owner.numberOfShares * 10;
+    const progressbarPink = '#cd6795';
+    const progressbarGray = '#F4F1F24';
+    const progressbarText = '30px';
```
dfangshuo: These should be capitalized to denote the fact that their constants:

e.g. PROGRESS_BAR_PINK

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
 
 class Investment extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      showModal: false,
+      newIsReceivingDividends: props.owner.isReceivingDividends
+    };
+
+    this.handleOpenModal = this.handleOpenModal.bind(this);
```
aivantg: We've switched away from using `bind` in our codebase. Use arrow functions instead! Fix is mentioned below

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+      isReceivingDividends: newIsReceivingDividends
+    });
+  };
+
+  onClickSavePreferences = async () => {
+    await this.onSavePreferencesPressed();
+    this.handleCloseModal();
+  };
+
+  handleOpenModal() {
```
aivantg: to get away with not having to bind this function and the 3 below it, switch this from 

`handleOpenModal() {` 

to 

`handleOpenModal = () => {`

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+import 'react-circular-progressbar/dist/styles.css';
+import { CircularProgressbar, buildStyles } from 'react-circular-progressbar';
+import { connect } from 'react-redux';
+import { updateOwner } from '../../lib/airtable/request';
+import { refreshUserData } from '../../lib/userDataUtils';
 import '../../styles/Investments.css';
+import GreenCheck from '../../assets/green_check.png';
+import RedX from '../../assets/red_x.png';
 
 class Investment extends React.PureComponent {
```
aivantg: Since this component now has state, you don't want to extend `PureComponent`, just `Component`

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      showModal: false,
+      newIsReceivingDividends: props.owner.isReceivingDividends
```
aivantg: this is good! But the issue here, is if props CHANGES, state won't update because the constructor is only called once. Instead we want to put this in 2 lifecycle methods: `componentDidMount` and `componentDidUpdate`. Frankly this is an edge case we can handle later, but I want to make the comment for the record

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+    const progressbarPink = '#cd6795';
+    const progressbarGray = '#F4F1F24';
+    const progressbarText = '30px';
+
+    const circularProgressBarStyles = {
+      pathColor: progressbarPink,
+      textSize: progressbarText,
+      textColor: progressbarPink,
+      trailColor: progressbarGray
+    };
```
aivantg: Could we move all these styles to the CircularProgressBar component?

aivantg: oh, i just realized that that wasn't a component we wrote. Something that would be good practice would be isolating the styles for the `CircularProgressBar` in our own component. Something that would look like this: 

You could pass the percentage value down as a prop!
```
import React from 'react';
import { CircularProgressbar, buildStyles } from 'react-circular-progressbar';

class SharesProgressBar extends React.Component { 
  render() {
    const styles = buildStyles(...)
    return (<CircularProgressBar .../>)
  }
}
```

src/screens/general/Investment.js
```diff
@@ -26,4 +209,8 @@ class Investment extends React.PureComponent {
   }
 }
 
-export default Investment;
+const mapStateToProps = state => ({
+  owner: state.userData.owner
```
aivantg: 😍

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+                </div>
+                <div className="dividends-button">
+                  <button
+                    type="button"
+                    className="change-button"
+                    onClick={this.handleOpenModal}
+                  >
+                    Change
+                  </button>
+                  <Modal
```
aivantg: As this page gets more complicated, we might want to move this modal to it's own component, but again, something we can tackle in the future

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+  };
+
+  onSavePreferencesPressed = async () => {
+    const { newIsReceivingDividends } = this.state;
+    await this.submitPreference({
+      isReceivingDividends: newIsReceivingDividends
+    });
+  };
+
+  onClickSavePreferences = async () => {
```
aivantg: this function seems kinda extraneous, why not put all of this in `onSavePreferencesPressed`?

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+    this.state = {
+      showModal: false,
+      newIsReceivingDividends: props.owner.isReceivingDividends
+    };
+
+    this.handleOpenModal = this.handleOpenModal.bind(this);
+    this.handleCloseModal = this.handleCloseModal.bind(this);
+  }
+
+  submitPreference = async newIsReceivingDividends => {
```
aivantg: I think we could consolidate some of these functions!

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+    const { owner } = this.props;
+    await updateOwner(owner.id, newIsReceivingDividends);
+    await refreshUserData(owner.id, owner.isReceivingDividends);
```
aivantg: Great use of redux!! `refreshUserData` doesn't need to take in the `owner.isReceivingDividends` part though, just the `ownerId`!

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+                    text={owner.numberOfShares}
+                    styles={buildStyles(circularProgressBarStyles)}
+                  />
+                </div>
+                <div className="box-text">
+                  <h5>
+                    You currently own {owner.numberOfShares} out of 10 possible
+                    shares
+                    <br />
+                    <h4>${owner.numberOfShares * 100}.00</h4>
```
aivantg: don't nest h4 within h5! I think you can just move it below to get the same effect?

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+                    </a>
+                  </div>
+                  <div className="investments-dividend">Divest</div>
+                </div>
+              </div>
+              <div className="investments-box-dividends">
+                <div className="dividends-preferences-box">
+                  <h4>Dividend Preferences</h4>
+                  <div className="status">
+                    <img
```
aivantg: no idea why...but when you navigate away from this page and then back to it, the size of the check doubles 😅

Maybe we could make the size fixed in the CSS class?

src/screens/general/Investment.js
aivantg: this issue is a thing, but it's not as important in this component. The main note is if the user refreshes while on this page, the value of `props.owner.isReceivingDividends` is actually undefined since the owner hasn't been loaded yet. We can talk about this later, but luckily it doesn't have too many implications for the code :)

src/screens/general/Investment.js
```diff
@@ -1,17 +1,200 @@
...
+                        <div>
+                          <input
+                            type="radio"
+                            id="yes-dividends"
+                            name="dividends-choice"
+                            value="yes"
+                            checked={
+                              newIsReceivingDividends ? 'checked' : false
+                            }
+                            onClick={() => this.handleCheckYes()}
```
aivantg: just turn handleCheckYes (and handleCheckNo) into arrow functions, and turn this into `onClick={this.handleCheckYes}`!

src/screens/general/Investment.js
cindyzhang12: - note: make new file for this

PR #111, opened 2020-03-16
Added Hi-Fi Loading Spinner

aivantg:

## What's new in this PR?

I Implemented the loading spinner Iris designed!

### Screenshots

loading

CC: @dfangshuo

### DIFF
--- a/src/components/PPLoadingSpinner.js
+++ b/src/components/PPLoadingSpinner.js
@@ -4,11 +4,11 @@ import '../styles/PPLoadingSpinner.css';
 class PPLoadingSpinner extends React.PureComponent {
   render() {
     return (
-      <div className="spinner">
-        <div className="dot blue-dot" />
-        <div className="dot pink-dot" />
-        <div className="dot yellow-dot" />
-        <div className="dot green-dot" />
+      <div className="pp-loading-spinner">
+        <div className="pp-loading-dot pp-loading-blue-dot" />
+        <div className="pp-loading-dot pp-loading-pink-dot" />
+        <div className="pp-loading-dot pp-loading-yellow-dot" />
+        <div className="pp-loading-dot pp-loading-green-dot" />
       </div>
     );
   }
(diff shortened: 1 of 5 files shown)


src/components/PPLoadingSpinner.js
```diff
@@ -0,0 +1,17 @@
...
+import '../styles/PPLoadingSpinner.css';
+
+class PPLoadingSpinner extends React.PureComponent {
+  render() {
+    return (
+      <div className="spinner">
+        <div className="dot blue-dot" />
+        <div className="dot pink-dot" />
+        <div className="dot yellow-dot" />
+        <div className="dot green-dot" />
```
dfangshuo: maybe name this something more specific like loading-...-dot etc. Am really trying to minimize style collisions now that our app is becoming rather large

dfangshuo: also nit: is this preferable to simply writing PPLoadingSpinner as a pure function since it doesn't hold any state?

src/styles/PPLoadingSpinner.css
```diff
@@ -0,0 +1,101 @@
...
+  animation: sk-green-dot-rotate var(--pp-loading-spinner-speed) infinite
+    ease-in-out;
+}
+
+.dot {
+  width: 20px;
+  height: 20px;
+  border-radius: 10px;
+  position: relative;
+}
```
dfangshuo: same comment as above with naming specificity. with `dot`, especially, maybe pre-appending loading to all

src/components/PPLoadingSpinner.js
aivantg: sure! 

Also we definitely could.....honestly in another world we'd have shifted towards functional, but at some point I arbitrary felt like we'd be better off by making everything class and keeping it consistent looking for someone new looking at our code. So "Functional Components" are denoted by `React.PureComponent`. We could change it! We should just change it everywhere

PR #112, opened 2020-03-16
Added Cancel Button to User Profile Page

aivantg:

## What's new in this PR?

Added a cancel button to user profile page.

### Screenshots

image

CC: @dfangshuo

### DIFF
--- a/src/screens/shared/UserProfile.js
+++ b/src/screens/shared/UserProfile.js
@@ -31,11 +31,11 @@ class UserProfile extends React.Component {
   componentDidUpdate(prevProps) {
     const { owner } = this.props;
     if (owner !== prevProps.owner) {
-      this.pullPropsIntoState();
+      this.pullPropsIntoState('both');
     }
   }
 
-  pullPropsIntoState = () => {
+  pullPropsIntoState = type => {
     const { owner, isLoadingUserData } = this.props;
 
     // If data isn't in redux yet, don't do anything.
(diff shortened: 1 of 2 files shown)


src/styles/UserProfilePage.css
```diff
@@ -55,6 +55,24 @@
...
+
+.contact-form-header-buttons {
+  display: flex;
+  flex-direction: row;
+}
+
+.contact-form-header-buttons button {
+  margin: 8px;
+}
+
```
dfangshuo: maybe pre-append userProfile to this to increase specificity?

src/screens/shared/UserProfile.js
```diff
@@ -31,28 +31,39 @@ class UserProfile extends React.Component {
   componentDidUpdate(prevProps) {
     const { owner } = this.props;
     if (owner !== prevProps.owner) {
-      this.pullPropsIntoState();
+      this.pullPropsIntoState('both');
     }
   }
 
-  pullPropsIntoState = () => {
+  pullPropsIntoState = type => {
```
dfangshuo: nit: im not sure what i feel about this name, as i feel like it should be more specific? and if it's this general, perhaps it could be in a more general-purpose file?

aivantg: It's not really a general purpose file...like an alternate name we use elsewhere is `refreshState` but there's no other way to describe what's happening 😅 It's just pulling the new props into state.

src/styles/UserProfilePage.css
aivantg: okie

PR #117, opened 2020-03-27
nick/dropdown

niwong:
## What's new in this PR?
IT HAS BEEN MADE PRETTY PRETTY. 

Created a basic dropdown navigational item used when hovering over the gear icon on any page.

Had to make changes in `app.js` and `NavBar.js` in order to pass down the `history` prop so that I could perform logic to log the user out properly. On the CSS side, I think everything is realigned (possibly even more than it was before) because I set every list element and the logo element in the NavBar to `display:flex`. This change aligned everything vertically. I used some CSS magic to get that cog positioned right. I honestly don't fully understand how Ashley positioned the dropdown, but I fiddled with the numbers and it looks good on my local machine.

### How to Review

Pull from `nick/dropdown`, run, and checkout `SettingsDropdown.js` in the highest level `components` folder.

### Related PRs

### Next steps

The CSS is an absolute (hahah, get it?) pain to align everything in a way that wasn't obtrusive to the existing styles. I need to get back to that and test on different screen sizes.

Also, when the user selects "Logout" from my dropdown, the `history` prop isn't passed to the parent component of my dropdown, namely `NavBar.js`, so I need to update that to correctly pass in props.

### Design Status

Preview: https://deploy-preview-117--peoplepower.netlify.com/

### Screenshots

CC: @dfangshuo


dfangshuo:
Is this ready for review? I'm not too sure cause you requested a review but it's in Draft? Mark it as ready for review once it is so it's clearer!

src/styles/NavBar.css
```diff
@@ -45,14 +46,92 @@ span {
...
+  height: 40px;
+  width: 40px;
+  position: relative;
+}
+
+.settings-dropdown .dropdown-content {
+  visibility: hidden;
+  background-color: white;
+  padding: 0.7rem 0.7rem;
+  color: #4f4f4f;
```
dfangshuo, line 85: use colors from main.css

src/styles/NavBar.css
```diff
@@ -45,14 +46,92 @@ span {
...
+  opacity: 0.8;
+}
+
+.dropdown-logout {
+  padding: 0px;
+}
+
+/* dropdown arrow */
+.settings-dropdown .dropdown-content::after {
+  content: '';
```
dfangshuo, line 123: hmm any reason why we need this line?

aivantg:
Okay this is another issue that i just discovered after playing with it, it's like...pretty hard to hover over settings and click something on the dropdown without it disappearing. You have to be _very_ specific about where your mouse is, which might not be a super great UX :/ 

Don't know if there's an easy way to change that (like a delay before it disappears or something? change the bound of it?) or if we need to change the UX. this can't just be an us problem..

niwong:
> Don't know if there's an easy way to change that (like a delay before it disappears or something? change the bound of it?) or if we need to change the UX. this can't just be an us problem..

Ahhh, I was running into that too. I thought increasing the height of the element that triggers the hover, and I thought it worked, but it seems that I'll need to make it bigger so the dropdown stays open.

src/styles/NavBar.css
niwong: Yes, this is what adds the little arrow to the dropdown.

PR #123, opened 2020-04-04
404 Error Page

cindyzhang12:

## What's new in this PR?

- created error page that shows up when someone enters an url that doesn't route to any page!
### How to Review

- check App.js for implementation of rerouting to the 404 page when an unidentifiable url is entered
- ErrorPage.js and ErrorPage.css for actual component and styling
### Related PRs

N/A
### Next steps
- talked to Aivant about this, but not sure what to use for styling in terms of px/vh/% and tried to use vh/% so it's more consistent across screens, please let me know if it's okay, if I should continue doing this in the future, if I should use px instead, etc.!
PR/has to be done "

### DESIGN STATUS

https://deploy-preview-123--peoplepower.netlify.com/

### Screenshots

CC: @dfangshuo


src/styles/ErrorPage.css
```diff
@@ -0,0 +1,52 @@
...
+
+.back-to-homepage-button {
+  width: 100%;
+  display: flex;
+  justify-content: center;
+}
+
+.back-to-homepage-button a {
+  background-color: var(--pp-blue);
+  border-radius: 4vh;
```
aivantg, line 44: woah, using vh for border-radius very cool

src/styles/ErrorPage.css
```diff
@@ -0,0 +1,52 @@
...
+  color: white;
+  text-align: center;
+  text-decoration: none;
+  display: inline-block;
+  font-size: 2.5vh;
+  width: 25vh;
+  height: 4vh;
```
aivantg, line 51: I think the text isn't centering well, because it's not centered _vertically_

I think maybe one way to fix this is to actually replace the `width` and `height` values with a `padding` value. 

When you don't specify a `width` or `height`, they default to being as small as they need to wrap the content (in this case, the text). then if you specify a `padding`, it'll give an equal space all around the button.

Just play with it until it looks good!

src/styles/ErrorPage.css
aivantg: I'd recommend using `100vh` to get a "perfect radius". This basically makes the border radius as large as it can possible be (100% of the view height)

src/styles/ErrorPage.css
```diff
@@ -0,0 +1,52 @@
...
+
+.back-to-homepage-button a {
+  background-color: var(--pp-blue);
+  border-radius: 4vh;
+  color: white;
+  text-align: center;
+  text-decoration: none;
+  display: inline-block;
+  font-size: 2.5vh;
+  width: 25vh;
```
aivantg, line 50: heads up, `vh`, which stands for "view height" has an analog called "vw" which is "view width". If you're trying to vary things based on the page's width, use that instead (don't need to change it here, just a note)

PR #139, opened 2020-04-16
nick/payment-success-screen

niwong:

## What's new in this PR?

Implemented mid-fi payment success screen for general owners buying shares and subscribers paying bills. 

### How to Review

Make a payment! Either pay a bill or buy a share.

Checkout `billPayment.js`, `BuyShares.js` and the new component I made `PaymentSuccessModal` in the shared components modal. I also changed a redux refreshState method in lib with the help of Aivant.

### Related PRs

### Next steps

PR/has to be done "

### Screenshots

### Design Status

CC: @dfangshuo


src/lib/userDataUtils.js
```diff
@@ -18,10 +18,12 @@ import {
 import { getCredentials } from './credentials';
 
 // Function takes in an ownerId and fetches the latest owner object and all associated user data
-const refreshUserData = async ownerId => {
-  // Save loading status to Redux
-  store.dispatch(setLoadingForUserData());
-  store.dispatch(setLoadingForAnnouncements());
+const refreshUserData = async (ownerId, loadSilently = false) => {
```
dfangshuo, line 21: is tthere ar eason loadSilently is set to false here?

src/screens/subscriber/components/BillingPayment.js
```diff
@@ -49,15 +51,24 @@ class BillingPayment extends React.Component {
 
   onPaymentSuccess = async (details, data) => {
     const { activeBill } = this.state;
-    const { history, owner } = this.props;
-    this.setState({ loadingPayment: false, loading: true });
+    const { owner } = this.props;
+    this.setState({
+      loadingPayment: false,
+      transactionAmount: details.purchase_units[0].amount.value,
```
dfangshuo, line 57: same comment as above

src/screens/general/BuyShares.js
```diff
@@ -18,22 +19,29 @@ class BuyShares extends React.PureComponent {
...
   }
 
   onPaymentSuccess = async (details, data) => {
-    const { owner, history } = this.props;
+    const { owner } = this.props;
     const { sharesBuying } = this.state;
-    this.setState({ loading: true });
+
+    this.setState({
+      transactionAmount: details.purchase_units[0].amount.value,
```
dfangshuo, line 33: there's a potential to crash the app here if unexpected data is returned. I make an if check to check that

1. purchase_units is not null
2. purchase_units has length >= 1
if these conditions aren't met, add a console.warn statement

src/lib/userDataUtils.js
niwong: Yes, in order to present the user current data after they navigated out of the payment success screen, we still needed to call `refreshUserData` but without the page reloading functionality, so we created this loadSilently option. It still refreshes data, but it doesn't force the page to reload.

src/screens/general/BuyShares.js
niwong: I added the check, but can we have an OnPaymentSuccess with a purchase of zero shares?

src/lib/userDataUtils.js
dfangshuo: sounds good! @aivantg ++ for great APIs

src/screens/general/BuyShares.js
dfangshuo: that is a great point, and i think points down to some design/architectural issues.

In general, I do not think it's a good idea to have to implicitly depend on the fact that the caller of the function will pass this function well-formed input. If we want to go with that pattern, we should enforce this pattern explicitly assertion statements + try/catches.

src/screens/shared/components/PaymentSuccessModal.js
```diff
@@ -0,0 +1,62 @@
...
+        >
+          <circle
+            cx="41.5"
+            cy="41.5"
+            r="39"
+            stroke="#CD6795"
+            strokeWidth="5"
+          />
+          <path
+            d="M26.7522 41.7906C25.7674 40.8229 24.1845 40.8367 23.2168 41.8216C22.2491 42.8064 22.263 44.3893 23.2478 45.357L26.7522 41.7906ZM34.593 53L32.8408 54.7832L34.593 56.5049L36.3452 54.7832L34.593 53ZM59.7522 31.7832C60.737 30.8155 60.7509 29.2326 59.7832 28.2478C58.8155 27.263 57.2326 27.2491 56.2478 28.2168L59.7522 31.7832ZM23.2478 45.357L32.8408 54.7832L36.3452 51.2168L26.7522 41.7906L23.2478 45.357ZM36.3452 54.7832L59.7522 31.7832L56.2478 28.2168L32.8408 51.2168L36.3452 54.7832Z"
```
dfangshuo, line 28: what is this :o

src/screens/shared/components/PaymentSuccessModal.js
```diff
@@ -0,0 +1,62 @@
...
+            fill="#CD6795"
```
dfangshuo, line 29: declare this as a constant at the top of the file

PR #154, opened 2020-04-25
Charts Data

aivantg:

## What's new in this PR?

- Populates the Charts with Data From Airtable
- Production Chart pulls from "Monthly Production Data" field on Solar projects
- Effective Cost chart pulls from would be cost + real cost data using same calculations as pp-node

### How to Review

### Related PRs

### Next steps

PR/has to be done "

### Screenshots

### Design Status

CC: @dfangshuo


src/components/ProductionChart.js
```diff
@@ -0,0 +1,179 @@
+/* eslint-disable class-methods-use-this */
```
dfangshuo, line 1: could we keep the name ProductionChartGeneral? Or name it something else? Right now its named the same thing as lib/charts/productionChart

src/components/ProductionChart.js
```diff
@@ -0,0 +1,179 @@
...
+const colors = ['pink', 'green', 'yellow'];
+
+class ProductionChart extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      equivalencyIndex: 0
+    };
+    this.style = !props.subscriberVersion ? '' : '-sub';
+    this.data = this.calculateSolarProjectProduction(props.solarProjects);
```
dfangshuo, line 27: interesting - is there a reason why these aren't in state?

src/components/ProductionChart.js
```diff
@@ -0,0 +1,179 @@
...
+    );
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.interval);
+  }
+
+  calculateSolarProjectProduction(solarProjects) {
+    // Combine Solar Project Production Data
+    const productionDataObject = solarProjects.reduce((d, project) => {
```
dfangshuo, line 46: maybe change d to something more self-explanatory haha

src/components/ProductionChart.js
```diff
@@ -0,0 +1,179 @@
...
+  SwitchTransition,
+  TransitionGroup
+} from 'react-transition-group';
+import { connect } from 'react-redux';
+import moment from 'moment';
+import '../styles/Equivalencies.css';
+import createProductionChart from '../lib/charts/productionChart';
+import Car from '../assets/Car-Icon-Square.png';
+import Trash from '../assets/Trash-Icon-Square.png';
+import Coal from '../assets/Coal-Icon-Square.png';
```
dfangshuo, line 16: Maybe append Icon to these

dfangshuo:
```suggestion
import CarIcon from '../assets/Car-Icon-Square.png';
import TrashIcon from '../assets/Trash-Icon-Square.png';
import CoalIcon from '../assets/Coal-Icon-Square.png';
```

src/components/ProductionChart.js
```diff
@@ -0,0 +1,179 @@
...
+  renderEquivalencyThree = () => (
+    <div className={`prod-equivalencies-right-inner${this.style}`}>
+      <div className={`prod-equivalencies-text${this.style}`}>
+        greenhouse gas emissions avoided by
+        <div className="prod-equivalencies-amt">38.8</div>
+        bags of waste recycled instead of landfilled
+      </div>
+      <img className="prod-equivalencies-icon" src={Trash} alt="fruit" />
+    </div>
+  );
```
dfangshuo, line 106: I think lib/charts/productionCharts would be a better place to put these, especially since to only has 1 default export rn?

src/components/ProductionChart.js
```diff
@@ -0,0 +1,179 @@
...
+            classNames="fade"
+            timeout={{ enter: 300, exit: 300 }}
+          >
+            {equivalencies[equivalencyIndex]}
+          </CSSTransition>
+        </SwitchTransition>
+      </div>
+    );
+  }
+
```
dfangshuo, line 132: ditto fcomment above

src/lib/subscriberUtils.js
```diff
@@ -33,12 +33,35 @@ const isBillPayment = payment => {
   return payment.type === BILL_PAYMENT_TYPE;
 };
 
+const getEffectiveCostData = async owner => {
+  const bills = (await getSubscriberBillsByIds(
+    owner.subscriberBillIds || []
+  )).filter(bill => bill.status !== 'Pending');
+  const round = (x, y = 2) => Number(parseFloat(x).toFixed(y));
```
dfangshuo, line 40: nit: we could even pull this out as a function of its own so we can use it nxt time for other stuff

src/components/ProductionChart.js
aivantg: yup can def rename to account for that

src/components/ProductionChart.js
aivantg: good...question. So i initially made these class variables cuz they didn't ever change and they don't need to trigger a re-render. They totally can live in state though, would probably be a little more consistent too

src/components/ProductionChart.js
aivantg: Nah, `lib/charts/productionCharts` (now renamed) is meant to solely export the HighCharts chart options. This file is the proper place for the UI of equivalencies to live (changed component name to reflect that too).

PR #170, opened 2020-04-30
Put dependencies in to their own folders

dfangshuo:

## What's new in this PR?

The goal of this PR is to put all dependencies into their own folders in `src/lib`, for ease of handoff and so its immediately clear what the dependencies are to Grayson

**ALSO removed semantic-UI, which we don't seem to be using from `package.json`

### How to Review

New Folder structure looks like this:

image

*Notable changes:*

- moved userDataUtils.js into `src/lib/redux` and renamed it to userData.js
- moved onboardingData into `src/screens/onboarding`
- No file directly import ANY dependencies now (Paypal, google maps etc.), instead (in some case), a dummy file exists in the appropriate folder in lib that just imports and exports a dependency, so the lib folder represents *the sole entry-point of external dependencies*.

For an example of what I mean by this, see `src/lib/highcharts/highcharts.js`

There has not been/should not have been changes in functionality - just look over the structure of the new codebase.

### Related PRs

### Next steps

PR/has to be done "

### Screenshots

### Design Status

CC: @dfangshuo


src/lib/paypal/paypal.js
```diff
@@ -50,4 +51,9 @@ const recordBillPayment = async (details, data, bill) => {
   });
 };
 
-export { calculatePaypalTransactionFee, recordBillPayment, recordSharePayment };
+export {
+  PayPalButton,
```
aivantg, line 55: iiiiinteresting pattern, have never thought about having one entrypoint for all external dependencies or if that helps or hurts code organizedness but, not against it ¯\_(ツ)_/¯

src/lib/highcharts/highCharts.js
```diff
@@ -0,0 +1,4 @@
+import Highcharts from 'highcharts';
+import HighchartsReact from 'highcharts-react-official';
+
+export { Highcharts, HighchartsReact };
```
aivantg, line 4: the only thing I'd say is that I can 100% see someone in the future just importing from the packages itself, so it seems like a level of control that's difficult to maintain with no concrete use case

PR #180, opened 2020-05-03
removed credentials from redux

aivantg:

## What's new in this PR?

removed credentials from redux. simple update, makes app a tad bit more secure and clean

### How to Review

just confirm things work, especially credential-related things

### Related PRs

### Next steps

PR/has to be done "

### Screenshots

### Design Status

CC: @dfangshuo

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -22,7 +22,8 @@ import {
   isSubscriberOwner,
   isSignedIn,
   Credentials,
-  isOnboarding
+  isOnboarding,
+  getCredentials
 } from './lib/credentials';
 import AuthenticatedRoute from './components/AuthenticatedRoute';
 import Investment from './screens/general/Investment';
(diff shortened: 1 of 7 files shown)


aivantg:
> looks good to me! so instead of storing a user's credential, as compute them as needed. Just for my own understanding, why is this more secure?

lol honestly, after i finished the PR I realized it wasn't that much more secure, this PR doesn't do much 

initial reasoning: if credentials is separated in it's own variable in redux, a user could just edit that variable in redux and make themselves an admin

now: they can still change their owner record locally and make themselves an admin LOL. if airlock works right, they wouldn't be able to do anything because of it but they could. 

main benefits - one source of truth for credentials, usually a good thing for clean code, it's slightly less obvious where the credentials live. Not a big deal, but i mean i already did it so LOL

